PR #17, opened 2015-09-17
Fix redis backend and add some doc

asmaps:
I changed some string/bytes handling in the Redis backend to make it work under python 3.4 - not sure if this broke some 2.x compatibility. Also I added a tiny bit of doc about the redis backend.


andrewgodwin:
Thanks! Nice catch on the docs, too.

PR #26, opened 2015-11-07
Wording changes to docs

maikhoepfel:
I eagerly read through the (excellent, thanks!) documentation for channels, and had to re-read one or two sentences. So being a good citizen, I'm suggesting a few fixes here and there.

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -11,7 +11,7 @@ which allow websites to communicate outside of this traditional cycle.
 
 And, beyond that, there are plenty of non-critical tasks that applications
 could easily offload until after a response as been sent - like saving things
-into a cache, sending emails or thumbnailing newly-uploaded images.
+into a cache or thumbnailing newly-uploaded images.
 
 Channels changes the way Django runs to be "event oriented" - rather than 
 just responding to requests, instead Django responses to a wide array of events
@@ -38,7 +38,7 @@ alternative is *at-least-once*, where normally one consumer gets the message
 but when things crash it's sent to more than one, which is not the trade-off
 we want.
 
-There are a couple of other limitations - messages must be JSON serialisable,
+There are a couple of other limitations - messages must be JSON serializable,
 and not be more than 1MB in size - but these are to make the whole thing
 practical, and not too important to think about up front.
 
@@ -121,7 +121,7 @@ However, the key here is that you can run code (and so send on channels) in
 response to any event - and that includes ones you create. You can trigger
 on model saves, on other incoming messages, or from code paths inside views
 and forms. That approach comes in handy for push-style
-code - where you use HTTP2's server-sent events or a WebSocket to notify
+code - where you use HTML5's server-sent events or a WebSocket to notify
 clients of changes in real time (messages in a chat, perhaps, or live updates
 in an admin as another user edits something).
 
@@ -239,7 +239,7 @@ start thinking about them. Remember, Django provides some channels
 but you're free to make and consume your own, and all channels are
 network-transparent.
 
-One thing channels do not, however, is guaranteeing delivery. If you need
+One thing channels do not, however, is guarantee delivery. If you need
 certainty that tasks will complete, use a system designed for this with 
 retries and persistence (e.g. Celery), or alternatively make a management
 command that checks for completion and re-submits a message to the channel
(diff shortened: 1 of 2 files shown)


docs/concepts.rst
```diff
@@ -5,13 +5,13 @@ Django's traditional view of the world revolves around requests and responses;
...
 
-That was fine when the internet was all driven by simple browser interactions,
+That was fine when the internet was driven by simple browser interactions,
 but the modern Web includes things like WebSockets and HTTP2 server push,
 which allow websites to communicate outside of this traditional cycle.
 
 And, beyond that, there are plenty of non-critical tasks that applications
 could easily offload until after a response as been sent - like saving things
-into a cache, or thumbnailing newly-uploaded images.
+into a cache, sending emails or thumbnailing newly-uploaded images.
```
andrewgodwin: I explicitly left off "sending emails" as people usually route this through a reliable system, and channels deliberately isn't. Perhaps modify to add `(with a retry mechanism)`?

docs/concepts.rst
```diff
@@ -120,22 +117,20 @@ In fact, this is how Channels works. The interface servers transform connections
...
-code - where you use HTTP2's server-sent events or a WebSocket to notify
-clients of changes in real time (messages in a chat, perhaps, or live updates
-in an admin as another user edits something).
-
 However, the key here is that you can run code (and so send on channels) in
 response to any event - and that includes ones you create. You can trigger
 on model saves, on other incoming messages, or from code paths inside views
-and forms.
+and forms. That approach comes in handy for push-style
+code - where you use HTTP2's server-sent events or a WebSocket to notify
```
andrewgodwin: HTTP2 server-side events do not actually allow arbitary notification of the client; they're more a way to push resources (images, css) to the browser before it requests them.

docs/concepts.rst
```diff
@@ -240,15 +235,15 @@ Next Steps
...
+start thinking about them. Remember, Django provides some channels
 but you're free to make and consume your own, and all channels are
 network-transparent.
 
-One thing channels are not, however, is guaranteed delivery. If you want tasks
-you're sure will complete, use a system designed for this with retries and
-persistence like Celery, or you'll need to make a management command that
-checks for completion and re-submits a message to the channel if nothing
-is completed (rolling your own retry logic, essentially).
+One thing channels do not, however, is guaranteeing delivery. If you need
```
andrewgodwin: Perhaps just "guarantee" here?

docs/getting-started.rst
```diff
@@ -233,7 +240,7 @@ Persisting Data
 Echoing messages is a nice simple example, but it's
 skirting around the real design pattern - persistent state for connections.
 Let's consider a basic chat site where a user requests a chat room upon initial
-connection, as part of the query string (e.g. ``http://host/websocket?room=abc``).
+connection, as part of the query string (e.g. ``https://host/websocket?room=abc``).
```
andrewgodwin: I was wrong originally - the protocol here should be `wss`

docs/getting-started.rst
```diff
@@ -451,7 +462,7 @@ command run via ``cron``. If we wanted to write a bot, too, we could put its
 listening logic inside the ``chat-messages`` consumer, as every message would
 pass through it.
 
-Linearization
+Linearisation
```
andrewgodwin: I've changed my mind on UK vs US English - if this is going to be merged into Django, it needs to use US English, so probably need to keep the z. Sorry!

docs/concepts.rst
maikhoepfel: Whoops, I meant to write that. Thanks.

docs/concepts.rst
maikhoepfel: That makes sense. I'll remove it to avoid scope creep for the PR. I do think it would be helpful to have a list of common task, and mentioning how well they could be achieved with channels.

docs/concepts.rst
maikhoepfel: Ah of course, you're right. But I just moved the sentence around, because I found it easier to understand that way. Did both of us mean [HTML5 server-side events](http://www.w3schools.com/html/html5_serversentevents.asp) instead of HTTP2?

docs/getting-started.rst
maikhoepfel: Fixed.

docs/getting-started.rst
maikhoepfel: No worries.

maikhoepfel:
Thanks for the quick review, and sorry for the late response. I pushed a new commit, hopefully addressing your comments. Could you have another look?

andrewgodwin:
Perfect, thanks!

PR #29, opened 2015-11-07
Run tests with tox and add flake8/isort

ekmartin:
Hi, great project! I started fixing the Python 3 errors you fixed earlier today yesterday, however I didn't manage to create a pull request before you had done it too.

I did some other changes too though, so here's a pull request for those, if you're interested:
* Added tox to run the tests for both Python 3.5 and Python 2.7 together with Django 1.6-1.9 (i.e. `tox -e py35-django-18`). 
* Added flake8 and isort using the Django settings found [here](https://github.com/django/django/blob/53ccffdb8c8e47a4d4304df453d8c79a9be295ab/setup.cfg) (`tox -e isort` and `tox -e flake8`).
* Fix an error where `Group.channels()` didn't actually return anything.

### DIFF
--- a/channels/backends/redis_py.py
+++ b/channels/backends/redis_py.py
@@ -1,6 +1,5 @@
 import time
 import json
-import datetime
 import math
 import redis
 import random
@@ -64,7 +63,11 @@ class RedisChannelBackend(BaseChannelBackend):
     def send(self, channel, message):
         # if channel is no str (=> bytes) convert it
         if not isinstance(channel, str):
-            channel = channel.decode('utf-8')
+            channel = channel.decode("utf-8")
+        # Write out message into expiring key (avoids big items in list)
+        # TODO: Use extended set, drop support for older redis?
+        key = self.prefix + uuid.uuid4().hex
+
         # Pick a connection to the right server - consistent for response
         # channels, random for normal channels
         if channel.startswith("!"):
@@ -72,9 +75,7 @@ class RedisChannelBackend(BaseChannelBackend):
             connection = self.connection(index)
         else:
             connection = self.connection(None)
-        # Write out message into expiring key (avoids big items in list)
-        # TODO: Use extended set, drop support for older redis?
-        key = self.prefix + uuid.uuid4().hex
+
         connection.set(
             key,
             json.dumps(message),
--- a/channels/backends/redis_py.py
+++ b/channels/backends/redis_py.py
@@ -1,11 +1,11 @@
-import time
+import binascii
 import json
 import math
-import redis
 import random
-import binascii
+import time
 import uuid
 
+import redis
 from django.utils import six
 
 from .base import BaseChannelBackend
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 dist/
 docs/_build
 __pycache__/
+.tox/
 *.swp
-
+*.pyc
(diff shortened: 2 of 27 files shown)


channels/backends/redis_py.py
```diff
@@ -64,7 +63,11 @@ def connection(self, index):
     def send(self, channel, message):
         # if channel is no str (=> bytes) convert it
         if not isinstance(channel, str):
-            channel = channel.decode('utf-8')
+            channel = channel.decode("utf-8")
+        # Write out message into expiring key (avoids big items in list)
+        # TODO: Use extended set, drop support for older redis?
+        key = self.prefix + uuid.uuid4().hex
```
ekmartin: I moved the definition of `key` up here, as it doesn't seem like it's defined when `index = self.consistent_hash(key)` is called here: https://github.com/andrewgodwin/channels/blob/master/channels/backends/redis_py.py#L71

ekmartin:
Any reason the session key is stored as bytes [here](https://github.com/andrewgodwin/channels/blob/master/channels%2Fdecorators.py#L54)? It seems to crash `session.exists()` since it tries concatenating a string with the byte-type session key. Changed it to a string [here](https://github.com/ekmartin/channels/commit/80a75fce85b285b0e1b063c1d81e6a5a90fe672f) and it seems to work.

channels/backends/redis_py.py
andrewgodwin: Yup, good catch. I only added this code a few days ago, so I'm clearly missing a test here.

andrewgodwin:
Couple of issues:

 * You've wrapped several lines that would easily fit within 119 characters (the Django line length standard) - I'm not sure why flake8 did this, since it's been told to use 119. There's also closing brackets on the end of the line, rather than on the following line at the original indentation level.
 * The test setup you've got here isn't great - the test suite is going to need a full host project and to run tests using the management commands inside that, which I've already committed some of to the repo.
 * Merge conflicts.

You're right that the session key could be a string, though. It would be nice if these were separate pull requests; there are some things here I would happily accept but it's all mixed in with other stuff and whitespace/code style cleanup too, which I like to keep separate.

ekmartin:
Thank you for the feedback!

* I've fixed the wrongly wrapped lines, so everything should be within 119 characters now. Also fixed the closing brackets on the end of the line issue.
* I'm probably missing something, but how would you run the tests without something similar to the `runtests.py` file I added [here](https://github.com/ekmartin/channels/blob/tests_and_py3/runtests.py#L10)? I thought that was generally how you did tests for reusable applications.
* I rebased in master, so there shouldn't be any conflicts now.

I'll submit a separate pull request for the string session key.

andrewgodwin:
You run the tests by going into the test project and running "manage.py test channels" - the main problem is that the tests need consumers to be assigned etc. and the hooks for that right now are in the app loading path.

It's not super important since when it gets merged into Django they'll become project-level tests and thus have access to a bit more flexibility. Still not sure how best to do the integration tests (even the current set require a running redis server to complete fully), but I'm sure I can come up with something.

ekmartin:
Ah, didn't know you wanted to use the testproject to run the tests, thought it was more of an example. Feel free to remove the tox and runtests files if you want to, but I thought it'd be useful to be able to run the tests with different Django/Python versions (`tox` now runs the tests with all combinations of Django 1.6 - 1.9 and Python 2.7/3.5).

PR #30, opened 2015-11-08
WIP Added logging

wengole:
@andrewgodwin So far we're not using Django logging per se, we're just setting up our own logger that outputs to STDOUT. I think this would be good for the default, but as we're running under Django, we should load the LOGGING dictionary as configuration, and if it's empty, default to what we have, so something is always logged.

I think that's what we discussed.

### DIFF
--- a/channels/log.py
+++ b/channels/log.py
@@ -1,0 +1,16 @@
+import logging
+
+
+def setup_logger(name, verbosity=1):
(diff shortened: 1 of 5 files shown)


andrewgodwin:
Yup, this looks good to me - feel free to push it up!

PR #34, opened 2015-11-18
Include cookies in request_info for websockets

ekmartin:
This adds a cookies field to the `request_info` dict, which should make https://github.com/andrewgodwin/channels/blob/master/channels%2Fdecorators.py#L110 work.


andrewgodwin:
Ah, thanks! I'll merge but be warned, I'm rewriting that entire module anyway to be a single HTTP/WebSocket interface :)

ekmartin:
No problem, sounds like a better solution!

PR #42, opened 2015-12-14
Changed three "these/those" into explicit names

reinout:
The original version was unclear to me. It sounded like Go channels were the ones that are network-transparent even though we mean to say that Django's are.

So I simply removed these/those and used "Go channels" and "Django channels" explicitly. (But *do* check whether I got it right! :-) )


andrewgodwin:
I think I agree; and yes, you got it right!

PR #45, opened 2015-12-15
updates runwsserver --accept-all in docs to runallserver

nim65s:
If I understand correctly, the doc is obsolete


andrewgodwin:
Nope, it's not ready yet, so is deliberately not in the docs. I might even just remove that whole section for now until it's more stable.

nim65s:
Ok, I'll wait, then, thanks :)

PR #46, opened 2015-12-15
send a serializable dict in py3

nim65s:
Not sure if runallserver should work already, but at least one needs to decode some bytes in PY3 before JSON serialization.


andrewgodwin:
It's not finished yet, and that bit is going to be changed around, so I'll close this and fix it separately.

PR #55, opened 2015-12-20
Documentation: running workers with uWSGI

bastianh:
Since there are sections about hosting Django with gunicorn and uWSGI in the main Django documentation I thought it might also be a good idea to show examples how to run channel worker / interfaces when using those.

Maybe there should be a seperate page for examples ?

### DIFF
--- a/docs/deploying.rst
+++ b/docs/deploying.rst
@@ -79,6 +79,49 @@ start dropping).
 
 TODO: We should probably ship some kind of latency measuring tooling.
 
+uWSGI-Mules
+^^^^^^^^^^^
+
+When you use uWSGI to run your django app it is also possible to let uWSGI
+take care of your workers.
+
+You would need to add a python script to your project that will act as a launcher,
+for example as `myproject/worker.py`::
+
+    import os
+
+    from channels import DEFAULT_CHANNEL_BACKEND
+    from channels import channel_backends
+    from channels.worker import Worker
+    from django.core.wsgi import get_wsgi_application
+
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "myproject.settings")
+
+
+    def run_worker():
+        # bootstrap application
+        get_wsgi_application()
+
+       channel_backend = channel_backends[DEFAULT_CHANNEL_BACKEND]
+
+       worker = Worker(channel_backend=channel_backend)
+       worker.run()
+
+
+    if __name__ == '__main__':
+       run_worker()
+
+
+you can then simply add as many `mules`_ you want to you uWSGI start options, for example
+with ``--mule=myproject/worker.py``.
+
+If you reload your uWSGI webworkers using signals or using the `master-fifo`_, uWSGI will also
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Yes, there should probably be a separate section for different deployment methods - a uWSGI one would need to cover deploying a WebSocket server too, though, since while mules can run the workers, they can't run arbitary socket listeners.

Do you know of a way to get uWSGI to hand over raw incoming socket connections to something that can handle them? It's possible that would be enough to pass off the rest of the work to a WebSocket implementation we already have, though at that point you're using uWSGI as little more than a process manager.

bastianh:
I have never done this, but uwsgi can do websockets and it has a routing system to route a path to a different server.

http://uwsgi-docs.readthedocs.org/en/latest/WebSockets.html

Until now I've always redirected websockets directly in nginx

bastianh:
Here is described how you can route requests to an other uwsgi or webserver. (http://uwsgi-docs.readthedocs.org/en/latest/InternalRouting.html)

But I guess most uwsgi users use nginx so an example conf with redirecting in nginx would also be nice.

andrewgodwin:
The uwsgi websockets module is unfortunately too high-level to easily plug into the existing code; someone would have to write code specifically from scratch to run on uwsgi's implementation. I'm sure someone will.

auvipy:
there is django-websocket-redis which use uWSGI webSocket in django project

nemesisdesign:
exactly, maybe here there's something interesting:
https://github.com/jrief/django-websocket-redis/blob/master/ws4redis/uwsgi_runserver.py

ZuSe:
+1

ssteinerx:
I'm building out a setup like this with Docker, but the configuration should be pretty similar.  I'm planning on asking for the whole thing to be added to the [channels-examples](https://github.com/andrewgodwin/channels-examples) repo when I'm done with it this week.  It goes from naked Django through a Dockerized deployment, step by step.

andrewgodwin:
Closing for now till we can get a full separate example with daphne running too.

eukreign:
@ssteinerx Any luck with your example? Would be useful to have.

ssteinerx:
@eukreign I had to go with a more traditional setup and, unfortunately, the Docker project had to be put aside for the moment.

ZuSe:
@andrewgodwin  Why would you close it? I really like the idea. It would avoid deploying one more component, especially for websites that are already running Django with uwsgi. Just leave it open until we find a way how to do it, it's kinda independent from daphne, isn't it?

andrewgodwin:
@ZuSe: It's closed because as far as I could tell it couldn't be done, as uWSGI does not expose a low-level enough API to integrate with Daphne (it would instead need native ASGI support, and that is something for the uwsgi project to tackle if they wish)

mostaszewski:
Has anyone managed to run Django Channels' websockets on uWSGI? I would like to avoid using Daphne and stay with the uWSGI.

PR #61, opened 2016-01-03
Draft proposal for Server Push messages

Lukasa:
This is basically a skeleton proposal of how server push messages could look. Happily, with the design of ASGI as written here they're fairly simple: the application basically just pushes a request to the protocol server, which encodes it appropriately into a PUSH_PROMISE frame and then starts handling the channel for the pushed response as normal. This maps pretty neatly onto the HTTP/2 stream logic as well.

At some point we may need to address whether this is allowed to *fail*, e.g. because the remote peer doesn't support pushed responses. For now, this is just an idea for how this might look.

### DIFF
--- a/docs/asgi.rst
+++ b/docs/asgi.rst
@@ -484,7 +484,7 @@ Channel: ``http.response.?``
 
 Keys:
 
-* ``status``: Integer HTTP status code. 
+* ``status``: Integer HTTP status code.
 
 * ``status_text``: Byte string HTTP reason-phrase, e.g. ``OK`` from ``200 OK``.
   Ignored for HTTP/2 clients. Optional, default should be based on ``status``
@@ -524,9 +524,16 @@ Keys:
 Server Push
 '''''''''''
 
-Send before any Response or Response Chunk. HTTP/2 only.
+Must be sent before any Response or Response Chunk messages.
 
-TODO
+Channel: ``http.response.?``
+
+Keys:
+
+* ``request``: A Request message. Both the ``body`` and ``body_channel`` fields
+  MUST be absent: bodies are not allowed on server-pushed requests. The
+  ``reply_channel`` set on this object will be used for all further messages
+  relating to the pushed response.
 
 
 WebSocket
--- a/docs/asgi.rst
+++ b/docs/asgi.rst
@@ -526,14 +526,32 @@ Server Push
 
 Must be sent before any Response or Response Chunk messages.
 
+When a server receives this message, it must treat the Request message in the
+``request`` field of the Server Push as though it were a new HTTP request being
+received from the network. A server may, if it chooses, apply all of its
+internal logic to handling this request (e.g. the server may want to try to
+satisfy the request from a cache). Regardless, if the server is unable to
+satisfy the request itself it must create a new ``http.response.?`` channel for
+the application to send the Response message on, fill that channel in on the
+``reply_channel`` field of the message, and then send the Request back to the
+application on the ``http.request`` channel.
+
+This approach limits the amount of knowledge the application has to have about
+pushed responses: they essentially appear to the application like a normal HTTP
+request, with the difference being that the application itself triggered the
+request.
+
+If the remote peer does not support server push, either because it's not a
+HTTP/2 peer or because SETTINGS_ENABLE_PUSH is set to 0, the server must do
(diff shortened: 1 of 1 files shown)


andrewgodwin:
I figured that server pushes should always soft-fail if unsupported - the HTTP/2 spec says the client can just drop them, anyway. Of course, it might be nice to communicate to the application that it's not wasting cycles and something is actually getting it; there's a reason I kind of want to have some sort of `supports_server_push` key, but it might get a bit verbose.

I do like the idea of just re-embedding the Request message, though. The reply channel thing is a little strange, though, not sure I like applications having to invent reply channels.

Maybe say that it should also contain a Response message embedded inside, and then have Server Push Chunks for larger bodies?

Lukasa:
> I figured that server pushes should always soft-fail if unsupported

Yeah, but what I was unsure on is how this kind of soft-failure is communicated. This actually applies to ASGI more generally: for example, if a TCP connection falls over while the application is sending data, how does the application find that out? Does the channel just get closed? I don't have a strong preference or suggestion here, but whatever you want to do here should be consistent, and we can use that method to notify about the failure of server push.

> The reply channel thing is a little strange, though, not sure I like applications having to invent reply channels.

That's fair enough, but I think a new channel does need to get created by something here: that way we keep a convenient isomorphism between HTTP/2 stream and channel. However, you've given me a much better idea:

What if, instead of the application creating the channel, the server did it? Specifically, what if, on receipt of a `Server Push` message, the application unpackages the `Request` sent in it and then treats it like a new request just received from the wire: creates the appropriate channels, fills them into the request object, and then sends the now-populated Request message back on the `http.request` channel? That means an application has to know relatively little about the server push: it 'pushes' a request for the resource, and from then on the application sees that as though it is entirely a real request.

This approach seems kind of neat to me, but I did come up with it, so what do you think?

andrewgodwin:
Well, ASGI in general doesn't have the concept of channel closure or message rejection; everything is sending data into the void with no idea about delivery, which is deliberate, since it keeps things simple and much easier to optimise and load balance. If a protocol server loses a connection, you get sent a message saying it's gone, but you can't just take a channel name and ask an ASGI layer if it's open or not, so it seems the only option is just to let applications send into the void.

Server push is limited to being part of an overall HTTP/2 request, anyway, so it's not like there's a long-running connection to be aware of. It's possible some kind of "channel is open" state could be built in, but that's more state that a channel layer would need to take care of, and I've been trying to avoid that as much as I can.

As for the idea of the protocol server re-injecting the request, that is quite a neat way to do it considering that a server push response should match whatever a server would usually send. My concern is ordering, though; you have to keep all messages tied to the same original request on the one reply channel, as if you spread them over multiple ones, relative ordering isn't guaranteed.

I think the approach you describe would make sense inside Django - if you want to use Server Push, you annotate the response from the view with URLs, and Django runs them through the view framework internally and gets responses - but having that as something ASGI protocol servers would have to implement seems like too much.

Maybe, give each Server Push message its own unique ID - kind of like the reply channel, but only unique within the overall response - and then have Responses and Response Chunks have that optional ID as a key, so the protocol server can tie them with the appropriate initial request (or, if it doesn't care about server push, just drop any messages with that key)

Lukasa:
>  My concern is ordering, though; you have to keep all messages tied to the same original request on the one reply channel, as if you spread them over multiple ones, relative ordering isn't guaranteed.

I don't follow: what's the ordering concern? From a H2 protocol level, the only bit of ordering that matters is when the PUSH_PROMISE frame is emitted, which the proposal satisfies. What ordering are you worried about?

andrewgodwin:
Oh, you're right, for some reason I thought the promise responses had to come before the main one, but the only ordering concern the HTTP2 spec says is promises-before-responses, which makes sense.

In that case, it's quite possible the protocol server could just re-inject requests on the exact same pathway.

Lukasa:
@andrewgodwin Would you like to see a draft that proposes that mechanism?

andrewgodwin:
Yes, if you have the time to write it up!

Lukasa:
Ok, that's done: how do you feel about the new content?

andrewgodwin:
Yes, I like that a lot. OK if I merge it in?

Lukasa:
Fine by me!

PR #71, opened 2016-02-06
Docs: reply_channel is a property of message

georgebrock:
The explanatory paragraph says:

> the response channel is a property (`reply_channel`) of the request

However, in the example, `reply_channel` was a global.

(I'm just reading through the channels docs for the first time, so it's possible I've got this wrong)

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -109,11 +109,11 @@ message. Suddenly, a view is merely another example of a consumer::
     # Listens on http.request
     def my_consumer(message):
         # Decode the request from JSON-compat to a full object
-        django_request = Request.decode(message.content)
+        django_request = Request.channel_decode(message.content)
         # Run view
         django_response = view(django_request)
         # Encode the response into JSON-compat format
-        message.reply_channel.send(django_response.encode())
+        message.reply_channel.send(django_response.channel_encode())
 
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Ah yes, the victim of my iterations there. In fact, not only is it now a property, but the property is an actual Channel object rather than a string you need to pass to `Channel()`. I'll fix it up tomorrow unless you want to take a stab!

georgebrock:
Even better. I've updated the PR.

georgebrock:
I noticed another inconsistency while looking for other outdated examples: this example used `Request.decode` and `response.encode` whereas other examples used `Request.channel_decode`, and `response.channel_encode`.

Even after this change, I'm not completely sure it's correct. It looks like `channel_encode` and `channel_decode` mostly disappeared in b9464ca149a4f66ebeb8801254cdd6d1b12cd58d, although they still show up in `channels/adapters.py` and some examples.

I'm possibly worrying too much about the docs anyway, given how much everything's still in flux.

andrewgodwin:
Yeah, I need to go through the docs again and rework them for how it's laid out now; I've been spending most of my time writing conformance test suites and examples to test the spec to make sure I don't need to change it again.

PR #99, opened 2016-03-25
Remove references to Django project name

chornsby:
This PR updates the "Getting started" docs to match the standard Django package layout, rather than importing local apps from the project package. 

While I was following the instructions I ran into the following configuration error:
```
django.core.exceptions.ImproperlyConfigured: Cannot import consumer 'myproject.myapp.consumers.http_consumer'
```
The more common project structure would have `myapp` at the same level as `myproject` so the import path would be just `myapp.consumers.http_consumer`.

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -64,7 +64,7 @@ Here's what that looks like::
     # In routing.py
     from channels.routing import route
     channel_routing = [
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Lovely, thanks!

PR #114, opened 2016-04-04
Remove lazy loading of middleware

evansd:
This mirrors the equivalent change in Django itsef. See:
https://code.djangoproject.com/ticket/26452
https://github.com/django/django/commit/99bb7fcc1859615a7b8c2468e7b97d54853bfb10

### DIFF
--- a/channels/handler.py
+++ b/channels/handler.py
@@ -7,7 +7,6 @@ import sys
 import time
 import traceback
 from io import BytesIO
-from threading import Lock
 
 from django import http
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Does this still work on Django 1.8? I'm not familiar with when the need to lazy-load it got removed.

evansd:
It does work on 1.8. I'm not exactly sure when the need to lazy-load got removed either. I think it was during the 1.7 app loading refactor (certainly eager loading works on 1.7) but it might possibly have been earlier.

andrewgodwin:
Good enough for me!

PR #118, opened 2016-04-05
Browser based basic test

benjaoming:
As discussed on IRC #django-channels, ~~here is a proof of concept that `Group.send` on inmemory backend does not work atm.~~

Recreating:

1. Go to `testproject`, `python manage runserver`
1. Navigate to the basic echo test: http://127.0.0.1:8000/ - it works! Reply received.
1. Navigate to the group test: http://127.0.0.1:8000/group/ - that also works!

### DIFF
--- a/testproject/chtest/consumers.py
+++ b/testproject/chtest/consumers.py
@@ -12,11 +12,16 @@ def ws_connect(message):
     logger.debug("Received ws connect")
     path = message.content['path']
     if path.endswith("group/"):
+        logger.debug("Connect request for group channel, adding to group channel")
         Group(TEST_GROUP).add(message.reply_channel)
 
 
 def ws_disconnect(message):
     logger.debug("Received ws disconnect")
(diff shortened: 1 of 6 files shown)


benjaoming:
I have now done testing with the Redis backend, and it's the same result: Replying directly back to a message works, but `Group.send` does not.

@andrewgodwin - could you have a look at my code and try running it and tell me if it works for you? This could be a problem related to my system (Ubuntu 14.04).

benjaoming:
Okay, although I don't understand the reasoning behind this behavior, nothing is sent when the message dict does not contain a `"text"` key. This has been the source of the problem all along.

Could we perhaps change to backends channel layers to fail loudly when this mandatory isn't present?

andrewgodwin:
The test project is actually only for benchmarking really, and so it's set up in a specific way to match the benchmarker script; I don't want to make it a general test thing (and really should rename it!)

PR #140, opened 2016-04-28
Update to clarify Python 3 compatibility

benjaoming:
Does this make sense? A fellow team member didn't believe that Channels were Python 3 ready after looking at the docs, so I though it's best to clarify it once and for all, for instance in the FAQ?


andrewgodwin:
Sure, best to have it clarified if people are unsure.

PR #146, opened 2016-05-08
Fix re-inserting, with tests

Krukov:


channels/tests/test_worker.py
```diff
@@ -35,3 +55,49 @@ def test_channel_filters(self):
...
+        def _consumer(message, **kwargs):
+            if not hasattr(_consumer, '_called'):
+                _consumer._called = True
+                raise ConsumeLater()
+
+        consumer = mock.Mock(side_effect=_consumer)
+        chenels = ['test', ]
+        channel_layer = mock.MagicMock()
+        channel_layer.router.channels = chenels
+        channel_layer.receive_many = mock.MagicMock(return_value=('test', {}))
```
andrewgodwin: You don't need to mock this; the ChannelTestCase base class provides a pre-made in-memory channel layer you can prep messages on and extract sent messages from in a much less fragile way. See https://github.com/andrewgodwin/channels/blob/master/channels/tests/test_request.py#L20

channels/tests/test_worker.py
```diff
@@ -35,3 +55,49 @@ def test_channel_filters(self):
...
+
+    def test_run_with_consume_later_error(self):
+
+        # consumer with ConsumeLater error at first call
+        def _consumer(message, **kwargs):
+            if not hasattr(_consumer, '_called'):
+                _consumer._called = True
+                raise ConsumeLater()
+
+        consumer = mock.Mock(side_effect=_consumer)
```
andrewgodwin: You seem to be using both mock.callcount and _consumer._called; is there a need for both? I'd recommend just combining call count into this inner part of the function and then you can drop the need to use mock and pass the consumer function in directly.

Krukov:
Is it better?  Also I added copying for routes at ChannelTestCase tests.

channels/worker.py
```diff
@@ -122,5 +122,7 @@ def run(self):
                         self.channel_layer.send(channel, content)
                     except self.channel_layer.ChannelFull:
                         time.sleep(0.05)
+                    else:
+                        break
```
andrewgodwin: Why is this here? It negates the for loop?

andrewgodwin:
Yes, that's better - thanks for working on this!

PR #159, opened 2016-05-18
Add opt '--jobs N' to start multiply workers.

onegreyonewhite:
This is a sketch. But it is workable.
I'm open to all sorts of tips.
This is a workable option, which may well be used.

Use as `./manage.py runworker --jobs 3` to start 3 threads.

PS .: I'm sorry my bad english.

### DIFF
--- a/channels/management/commands/runworker.py
+++ b/channels/management/commands/runworker.py
@@ -5,7 +5,8 @@ from django.core.management import BaseCommand, CommandError
 from channels import DEFAULT_CHANNEL_LAYER, channel_layers
 from channels.log import setup_logger
 from channels.worker import Worker
-import threading,time
+import threading
+import time
 
 workers = []
 
@@ -31,7 +32,6 @@ class Command(BaseCommand):
             '--jobs', action='store', dest='jobs', default=1, type=int,
             help='Multiple start workers.',
         )
-
 
     def handle(self, *args, **options):
         # Get the backend to use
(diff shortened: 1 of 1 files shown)


andrewgodwin:
I don't think this would make sense, as generally python threading is a lot slower than using multiple processes, so we should be encouraging people to start N copies of runworker, not add threads (which might even make performance worse, based on some of my tests).

Going to have to reject this... sorry!

mrasband:
As an alternative @onegreyonewhite you could use [`circus`](https://circus.readthedocs.io/en/latest/) (or supervisor, etc.):

```
[watcher:interfaceserver]
cmd = daphne
arguments = ...

[watcher:workers]
cmd = python
arguments = manage.py runworker
num_processes = 3
```

This has the added benefit of restarting a worker (or whatever is watched) if it crashes for some reason.

onegreyonewhite:
It`s too hard, to fast deploy, @mrasband
I think about [pools](https://docs.python.org/2/library/multiprocessing.html#using-a-pool-of-workers). How do you like this idea?

oTree-org:
@onegreyonewhite you could also implement a management command (e.g. ``manage.py runworkers``) that uses Honcho to start processes like this:

```
from django.core.management.base import BaseCommand
from honcho.manager import Manager
import os
import sys

class Command(BaseCommand):
    help = 'Run N channels workers'

    def add_arguments(self, parser):
        super(Command, self).add_arguments(parser)
        parser.add_argument(
            '-n', '--num-workers', action='store', dest='num_workers', default=6,
            type=int,
            help='Number of worker processes')

    def handle(self, *args, **options):
        manager = Manager()

        for i in range(options['num_workers']):
            manager.add_process(
                'worker{}'.format(i),
                'python manage.py runworker',
                quiet=False,
                env=os.environ.copy())

        manager.loop()
        sys.exit(manager.returncode)
```

onegreyonewhite:
@oTree-org thank you for your solution.

PR #162, opened 2016-05-18
More tests utils for happy users

Krukov:
Now, if you need to test channels at the real application you need to know how the channels works (with some internal details).
There are some new test utils to help user with that problem, it is just proof of concept.
`Client` class - like `django Client`, allowing you to test your consumers in a easy way.
 
`apply_routes` provides ability to test your consumers atomic, without side effects (I think It is useful for third-party packages).
 
You can see simple example here https://github.com/Krukov/channels-examples/blob/example-new-tests-utils/multichat/chat/tests.py

@andrewgodwin What do you think about this?


andrewgodwin:
I like this, though it looks to be specifically a WebSocket/HTTP test client (route override is great though) - can you split out the client into something like `tests/http.py`, or have a base class without the cookie/session/login stuff and a subclass with that baked in?

Krukov:
OMG... Flake8 .. :heart: . Yes, I have tried. Please review it.

andrewgodwin:
Yes, Flake8 is really picky, haha! Looks good - there's a couple of things that might need to be changed in future, but those can happen in master. Merging.

PR #163, opened 2016-05-19
Remove inherited methods and relative imports

Krukov:
Sorry. I  forgot to remove it after copy/paste. :fire:


andrewgodwin:
Haha, I was going to dig in there later to see if I could trim it down!

PR #171, opened 2016-05-28
In-between method for adding decorators in right way + a few tests for generic consumer

Krukov:
At the current implementation `enforce ordering` decorator call with the result of the dispatched method..

### DIFF
--- a/channels/generic/base.py
+++ b/channels/generic/base.py
@@ -6,11 +6,11 @@ class BaseConsumer(object):
     Base class-based consumer class. Provides the mechanisms to be a direct
     routing object and a few other things.
 
-    Class-based consumers should be used directly in routing with their
-    filters, like so::
+    Class-based consumers should be used with route_class in routing, like so::
 
+        from channels import route_class
         routing = [
-            JsonWebsocketConsumer(path=r"^/liveblog/(?P<slug>[^/]+)/"),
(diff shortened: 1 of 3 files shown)


andrewgodwin:
So, I like everything in this patch apart from the addition of the `get_handler` method - what does it do separate from `dispatch`? I'd rather not add another function call and indirection in there without good reason, and `dispatch` was meant to be the nice overrideable thing people can use to change how lookups work.

Krukov:
I thought, that it helps to wrap "consumers" at different decorators in more clear way, just with calling super method with all args and kwargs. like this:
```python
def get_handler(self, message, **kwargs):
    return decorator(super().get_handler(message, **kwargs)
``` 
But now if you want to add some decorator you need to **get** super, **wrap** it and **call** , like this: 
```python
def dispatch(self, message, **kwargs):
    return decorator(super().dispatch)(message, **kwargs)
```
More familiar way - first of all **call** super method and than **wrap** it. So you did: 
```enforce_ordering(super(WebsocketConsumer, self).dispatch(message, **kwargs), slight=False)```
So it is wrong, but more familiar. `get_handler` is approach to solving problems.

andrewgodwin:
Ah, your point about wrapping the existing fetch logic with more decorators is a sensible one; I'm still not super keen on having two places people can override the dispatch logic, but I guess we can point them to change `get_handler` and leave `dispatch` alone.

Issue #176, opened 2016-05-29
authentication decorators not working?


DebVortex:
Hi!

I'm currently trying out channels and following the instructions about [authentication](http://channels.readthedocs.io/en/stable/getting-started.html#authentication) in the docs and I seem to stumble upon a problem.

Right now, I have a view that is uses the [LoginRequiredMixin](https://django-braces.readthedocs.io/en/latest/access.html#loginrequiredmixin) from braces. So once I'm logged in, I can access the View that will take care of connecting to the websockets.

My ``websocket.connect`` consumer looks like the following:

```
@channel_session_user_from_http
def ws_connect(message):
    if message.user.is_authenticated():
        Group(message.user.username).add(message.reply_channel)
        Group(message.user.username).send({'text': 'Conected!'})
    else:
        message.reply_channel.send({'text': 'Not authenticated.'})
```

However, I always have a ``AnonymousUser`` instead of the authenticated user.

Do I do anything wrong or is there a problem regarding the ``channel_session_user_from_http`` decorator?

andrewgodwin:
The most common cause of this is the session cookie not making it through to the WebSocket; are you running both the site and the WebSocket through the same domain/port (if you're on runserver, then this is true by default).

You can see what cookies you have in `message['cookies']` - would be good if you can compare that to the ones you get in a normal view to make sure they're coming through alright.

DebVortex:
Hey  @andrewgodwin,

thanks for the quick response. I checked a min ago and ``message['cookies']`` raises a KeyError. By Accessing the ``__dict__``-Attribute, shows there is also nothing similar. An idea what the reason could be?

andrewgodwin:
My mistake, it's `message['headers']` - there's no decoded cookies attribute in ASGI!

DebVortex:
I have a headers key, however no cookie data in there.

DebVortex:
The problem sits in front of the screen! In my JS I used ``ws://127.0.0.1:8000/ws``. However, I acceded the page via ``http://localhost:8000``. Thanks anyway and please, keep up the good work

I will close this Issue now :)

andrewgodwin:
Aha! I have done that too. Will see if I can add it in our docs somewhere :)

ostcar:
Maybe the documentation could use 
```javascript
var socket = WebSocket('ws://' + location.host + '/ws');
```
instead of 
```javascript
var socket = WebSocket('ws://127.0.0.1/ws');
```

[...]

PR #178, opened 2016-05-29
Use window.location.host instead of 127.0.0.1

ostcar:
See #176


andrewgodwin:
Surely using `window.location.hostname` would be better?

andrewgodwin:
Ah, I think `window.location.host` might include the port?

ostcar:
sry. Typo. I meant ```location.host``` this contains the port. I always used ```location``` and never ```window.location``` I did not even know that ```window.location``` is the right object. I change it.

PR #207, opened 2016-06-14
Call socket.send instead of in the onopen handler.

tbeadle:
The websocket is already open by the time the onopen handler was getting
defined.  Instead, just call socket.send() directly.  That causes the alert
message to show up immediately, as intended.

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -138,9 +138,7 @@ need to change the socket address if you're using a development VM or similar)::
     socket.onmessage = function(e) {
         alert(e.data);
     }
-    socket.onopen = function() {
-        socket.send("hello world");
-    }
+    socket.send("hello world");
 
 You should see an alert come back immediately saying "hello world" - your
 message has round-tripped through the server and come back to trigger the alert.
@@ -232,9 +230,7 @@ code in the developer console as before::
     socket.onmessage = function(e) {
         alert(e.data);
     }
-    socket.onopen = function() {
-        socket.send("hello world");
-    }
+    socket.send("hello world");
 
 You should see an alert come back immediately saying "hello world" - but this
 time, you can open another tab and do the same there, and both tabs will


andrewgodwin:
Are you sure? It can take websockets a while to handshake and open; there's no guarantee it's opened by that point. Do you have an API spec source?

tbeadle:
Maybe it's just because I was typing it in manually and not copy/pasting it into my javascript console.  When I did that, using the onopen method generated no alert.

andrewgodwin:
Right, as when you're typing there's enough delay between the two lines that it can negotiate; as a script it runs very fast, and doing what you suggest won't work.

Maybe change the PR to add a comment saying that if you're typing it in, not pasting, that you should do it a different way? Or can we find a way that works both with typing and straight execution?

tbeadle:
You could do something like this:

```
    var func = function() {
        if (socket.readyState == 1) {
            socket.send('hello world');
        } else {
            setTimeout(func, 250);
        }
    };
    setTimeout(func, 250);
```

but I'm afraid that's detracting from the simplicity of the example.

tbeadle:
Actually, the last line could just be  `func()`.

andrewgodwin:
That does seem slightly complicated; I think it might be nicer to just have a short callout explaining what `onopen` means and how you might have to run the `send` by itself.

tbeadle:
Agreed.

ssteinerx:
I made a documentation change regarding this, but it seems to have evaporated.
When you're following the documentation, testing the initial connection in the JS console, the onopen() never gets called so nothing happens.
Manually calling socket.send('msg') makes the echo happen.
Obviously, in a script, this is going to be different, but in the 'getting started' docs, this should definitely be changed as the naive user will be left scratching their head not knowing to call socket.send to start the process once stalled.

andrewgodwin:
Closing this as I'm going to write a better solution that checks socket state.

PR #214, opened 2016-06-17
Test runserver

readevalprint:
This PR brings run the `runserver` and `runworker` management commands up to 91% and 95% respectively.  I tried to change as little in the code as possible but to capture the logging, I had to move the `handler = logging.StreamHandler()` line out of the function into the module so that it could be overridden. Without that I couldn't find a place to mock the logging.

### DIFF
--- a/channels/tests/test_management.py
+++ b/channels/tests/test_management.py
@@ -43,8 +43,8 @@ class RunWorkerTests(TestCase):
             mock_worker.assert_called_with(
                 only_channels=None, exclude_channels=None, callback=None, channel_layer=mock.ANY)
 
-            cl = mock_worker.call_args[1]['channel_layer']
-            static_consumer = cl.router.root.routing[0].consumer
+            channel_layer = mock_worker.call_args[1]['channel_layer']
+            static_consumer = channel_layer.router.root.routing[0].consumer
             self.assertIsInstance(static_consumer, StaticFilesConsumer)
 
     def test_runworker(self, mock_worker):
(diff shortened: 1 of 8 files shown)


channels/tests/test_management.py
```diff
@@ -0,0 +1,133 @@
...
+                                             host='localhost', action_logger=mock.ANY, channel_layer=mock.ANY)
+
+    @mock.patch('sys.stderr', new_callable=StringIO)
+    @mock.patch('sys.stdout', new_callable=StringIO)
+    @mock.patch('channels.management.commands.runserver.Server')
+    def test_runserver_noworker(self, mocked_server, mock_stdout, mock_stderr):
+        '''
+        No need to Mock the Worker because it should not be used.
+        '''
+        call_command('runserver', '--noreload', '--noworker')
```
andrewgodwin: What is this testing? You seem to have mocked out the command being tested so it'll vacuously be true.

andrewgodwin: There's no assert here - what are you testing?

andrewgodwin:
Seems fine overall, but I'd like some more comments/docstrings overall describing what's happening, as mocking this much is somewhat impenetrable to others at a first glance.

channels/tests/test_management.py
readevalprint: Oh wow I just saw your comment. Good catch, let me flesh this out, not sure what I was thinking.

readevalprint:
I'm only mocking the `sys.stdout/err` to suppress logging from the test output filling the UnitTest output. But if that is not important I can remove it. But I will add more comments regardless.

readevalprint:
@andrewgodwin Ok, added more comments and fleshed out the assertions more. Let me know what think.

andrewgodwin:
Sorry for the delay here - it's been a hell of a week. Looks good, thanks!

readevalprint:
No worries, thanks!

Issue #221, opened 2016-06-25
WebSocket class-based consumer should have a http_user class body option


pashinin:
Hi, I tried class based consumers, but `message.user` is always `AnonymousUser()`

```
class C(WebsocketConsumer):
    channel_session_user = True

    def connect(self, message, **kwargs):
        print(message.user)
```

I digged and it happens here: https://github.com/andrewgodwin/channels/blob/master/channels/auth.py#L40

Can you help with it?

andrewgodwin:
You need to make sure the session cookies are coming through to Daphne - how have you set up your site to run? Are the websockets and the website where you're logged in on the same domain and port (e.g. `localhost` is different to `127.0.0.1`)?

pashinin:
Whole test project is on daphne. Run with this:

```
daphne proj.asgi:channel_layer --port 8000 -b 0.0.0.0 -v2
```

plus a worker:

```
python manage.py runworker -v2
```

There were session cookies in `message.headers` when printed them if that's what needed.

andrewgodwin:
Is there a session present in `message.session`, as well? It's worth comparing the session keys of HTTP and WS responses to work out what the difference is.

andrewgodwin:
Err, make that `message.http_session`.

pashinin:
Will check now

pashinin:
That's what I get: no sessions (exceptions printed)
```
AnonymousUser
'Message' object has no attribute 'http_session'
'Message' object has no attribute 'session'
```

andrewgodwin:
Oh, right, you haven't done the handoff from http session to channel session with the user object - the WebSocket endpoint doesn't have that yet. Let me retitle the issue to reflect that.

andrewgodwin:
In the meantime, you need to wrap your connect in channel_session_user_from_http.

pashinin:
Yeah, I thought this function will appear here. Thanks for the project. Waiting for fix.

andrewgodwin:
Released in `0.16.1`.

pydanny:
After upgrading to 0.16.1 I'm still having a problem with anonymous users and generics. I've tried running it using the following:

``` bash
python manage.py runserver
```
and
``` bash
python manage.py runworker -v2
daphne proj.asgi:channel_layer --port 8000 -b 0.0.0.0 -v2
```

My code is pretty simple:

``` python
from channels.generic.websockets import JsonWebsocketConsumer

[...]

PR #227, opened 2016-06-28
#221: generic connect transfer user if needed

thewayiam:
Not so flexable, but keep the most common usage work.
Would you want to let user config each method_mapping, something like:
```
method_mapping = {
    "websocket.connect": "raw_connect",
    "websocket.receive": "raw_receive",
    "websocket.disconnect": "raw_disconnect",
}

channel_session_user = {
    "websocket.receive": True
    "websocket.disconnect": True,
}

channel_session_user_from_http = {
    "websocket.connect": True,
}
```
so user could write/config their own class-based consumers from scratch

### DIFF
--- a/channels/generic/base.py
+++ b/channels/generic/base.py
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
 from channels.sessions import channel_session
-from channels.auth import channel_session_user
+from channels.auth import channel_session_user, channel_session_user_from_http
 
 
 class BaseConsumer(object):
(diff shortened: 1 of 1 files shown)


andrewgodwin:
This needs to be added, but:

- In the Websocket base class, not the overall base class
- With a separate class body variable rather than using channel_session_user

So I'm going to reject this PR as-is - sorry! I'm working on getting it added (see #221), but you're welcome to try too.

thewayiam:
Thanks for the comment!

PR #233, opened 2016-07-03
typo one -> once in concepts.rst

tymofij:


ssteinerx:
"to more than once" is not proper English, this is not an improvement.

PR #255, opened 2016-07-20
Register Bindings as soon as they are declared

AlexejStukov:
With the old approach - keeping a list of classes and registering them in the ``ready``-method in ``apps.py`` - I had the problem that ready got called before my ``consumers.py``, where my bindings are, was loaded. So ``register_all`` ran on an empty list and my bindings where not registered.  I also added a uid to the signal-connections to prevent duplicate messages.

### DIFF
--- a/channels/apps.py
+++ b/channels/apps.py
@@ -1,8 +1,6 @@
 from django.apps import AppConfig
 from django.core.exceptions import ImproperlyConfigured
 
-from .binding.base import BindingMetaclass
-
 
 class ChannelsConfig(AppConfig):
 
--- a/channels/apps.py
+++ b/channels/apps.py
@@ -20,5 +20,3 @@ class ChannelsConfig(AppConfig):
         # Do django monkeypatches
         from .hacks import monkeypatch_django
         monkeypatch_django()
-        # Instantiate bindings
-        BindingMetaclass.register_all()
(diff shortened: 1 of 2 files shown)


andrewgodwin:
This isn't allowed under the Django import rules - you're not allowed to do things to models before ready() is called, and the register phase can take strings and resolve them to models, so I'm going to have to reject.

The fundamental problem is correct, but we should have documentation about where to put bindings (i.e. in a file that's imported from models.py, or in models.py itself)

AlexejStukov:
Oh, ok. But for me it was actually the problem that ready got called before the ``binding_classes``-list was populated.  Django also [advices agains](https://docs.djangoproject.com/en/1.9/topics/signals/)t signal registration in ``models.py``. Maybe use a separate file like ``bindings.py``?

andrewgodwin:
Well, the signals aren't registered until the register_all action gets called, so they're not actually registered in models.py; they're registered in the ready() method, which is the best place for them.

`bindings.py` would also work, but it would still need to be imported at project load time, which means either an import from it in models or in an `__init__.py`.

AlexejStukov:
Got your point. But models.py just does not feel right as a place for bindings. And a binding.py would probably be first included in the routing.py and that is already too late, as i understand, because ready was already run. How about a route_binding class/method that does the registration? Would have the benifit, that the bindings only will be registered, if they are routed to.

andrewgodwin:
The best thing is probably to have the register_all method set a flag on the class that tells future registrations to happen immediately.

AlexejStukov:
I've opened a new pull request #260 with a register_immediately flag.

PR #256, opened 2016-07-20
add a modelname to the payload dict

AlexejStukov:
Would be sometimes really useful, especially if you use one Multiplexer with several Bindings


andrewgodwin:
This actually used to be in there, but it used the appname.modelname variant, which I think is still around as `self.model_label` - do you mind moving it to use that?

AlexejStukov:
Shure, as long as there is a way to identify the models.

AlexejStukov:
I've implemented it in a new pull request  #258

PR #262, opened 2016-07-21
Added WebsocketBindingWithMembers

AlexejStukov:
WebsocketBindingWithMembers inherits WebsocketBinding and additionally enables sending of member variables, properties and methods.


AlexejStukov:
Might also be integrated in ``Binding`` or as a Mixin.

channels/binding/websockets.py
```diff
@@ -87,3 +87,42 @@ def update(self, pk, data):
...
+    """
+
+    send_members = []
+
+    def serialize_data(self, instance):
+        data = super(WebsocketBindingWithMembers, self).serialize_data(instance)
+        for m in self.send_members:
+            member = getattr(instance, m)
+            try:
+                data[m] = member()
```
andrewgodwin: I suggest using `if callable(member)` here instead?

channels/binding/websockets.py
```diff
@@ -87,3 +87,42 @@ def update(self, pk, data):
...
+            except TypeError:
+                data[m] = member
+        return data
```
andrewgodwin: Maybe consider catching/warning about types that won't serialize into JSON; for the fields, this is done automatically with the Django serializer.

AlexejStukov:
Wouldn't  the encoding step already raise a ``TypeError: XYZ is not JSON serializable``?

andrewgodwin:
Well the Django code specifically encodes a few extra types using known patterns, so I guess we could use that, but not required.

Also, absolute -1 to mixins. I will not merge with one :)

AlexejStukov:
Hm, ok; then undo the change. Why do you dislike them?

andrewgodwin:
They make it very hard to debug where things come from and are generally very unfriendly to newer Python programmers, so I tend to avoid them at all costs. I've had to maintain too many codebases that rely on mixins to work and it's impossible to work out where anything happens.

AlexejStukov:
Now it's implemented as a normal class again.

andrewgodwin:
Yup, nice.

PR #264, opened 2016-07-21
Security fix - every field of a model is send - even password

AlexejStukov:
Atm WebsocketBinding sends every field of a model, even the password of a user. Users of the class should have to think about which fields they want to send to the user. Also added a more intuitive option for sending all fields.


andrewgodwin:
Agreed - this change presumably errors if fields is not provided?

AlexejStukov:
Now it does check for empty fields

andrewgodwin:
Thanks - I'll fix up the docs to match!

PR #266, opened 2016-07-24
as_route method for class based consumers

Krukov:
Hello @andrewgodwin 

It is proof of concept. 
I hope that somebody creates a really generic Consumers, so at this way it will be real nice to have method that gives more fluent way to use CB consumers.
Old way:
```python 
# consumers.py
from mygeneric import ModelBaseConsumers
from .models import User

class UserConsumers(ModelBaseConsumers):
    model = User
    fields = ['username', 'group']

#routes
from channels import route_class
from .consumers import UserConsumers, ....

routes = [
   route_class(UserConsumers, path='/users/'),
   ...
]
```

New way:
```python
from mygeneric import ModelBaseConsumers
from .models import User, Message

routes = [
    ModelBaseConsumers.as_route(model=User, fields=['username', 'group'], path='/users'),
    ModelBaseConsumers.as_route(model=Message, fields=['user', 'text'], path='/messags'),
]
```

What do you think about this?

### DIFF
--- a/channels/generic/base.py
+++ b/channels/generic/base.py
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
-from channels.sessions import channel_session
-from channels.auth import channel_session_user
+from ..sessions import channel_session
+from ..auth import channel_session_user
 
 
 class BaseConsumer(object):
--- a/channels/generic/base.py
+++ b/channels/generic/base.py
@@ -39,16 +39,14 @@ class BaseConsumer(object):
         return set(cls.method_mapping.keys())
 
     @classmethod
-    def as_route(cls, **kwargs):
+    def as_route(cls, filters, **kwargs):
         """
-        Return the route_class that can directly using at a routes
+        Shortcut function to create route with filters to direct to a class-based consumer with given kwargs
         """
         _cls = cls
-        _kwargs = dict(kwargs)
-        _dict = {key: _kwargs.pop(key) for key in kwargs.keys() if key in dir(_cls)}
-        if _dict:
-            _cls = type(cls.__name__, (cls,), _dict)
-        return route_class(_cls, **_kwargs)
+        if kwargs:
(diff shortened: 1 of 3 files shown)


andrewgodwin:
The ability to pass kwargs is quite nice, but the implementation using `type()` is a little verbose. Do we really need to copy the dict like that or can the `kwargs` just be passed in as the class body?

Also, if there are no kwargs, maybe skip making the new class and just pass in `cls`?

Krukov:
@andrewgodwin I tried to fix it

Krukov:
Looks like something strange with flake8 and travis...

andrewgodwin:
That was odd!

I'm still a little conflicted about the mixing of class body settings and route settings in the same kwarg list, and which one they apply to being implicit (not to mention you wouldn't be able to set the `path` routing key if there was anything called `path` on the class). Do you have any ideas to deconflict them?

Krukov:
I suggest to using `filters` key word argument for filters and `kwargs`  for class body. Will it be more clear?

```python
routes = [
    ModelBaseConsumers.as_route(model=User, fields=['username', 'group'], filters={'path': '/users'}),
    ModelBaseConsumers.as_route(model=Message, fields=['user', 'text'], filters={'path': '/messags'}),
]
```

poiati:
Isn't `filters` bit problematic because of the `QuerySet#filter`, since we are dealing with the Model a thus the `fields` argument, this can be confusing. What about `match`?

andrewgodwin:
I actually might be more a fan of keeping the routing filters as the things outside as normal kwargs - to match `route()` calls - and moving the class body overrides into a named variable, maybe `attrs` or something.

Krukov:
@andrewgodwin  Suitable naming.

andrewgodwin:
Yes! I like that a lot. Sorry to ask for so many changes :)

Krukov:
No problems @andrewgodwin . Thanks you!

PR #267, opened 2016-07-24
Improvements, fix and tests

Krukov:
I wanted to write a couple of tests for new feature (bindings), and I did it. Does it help you?

PR contains breaks changes at `HttpClient.receive` method, is it critical?


andrewgodwin:
Given that it's not documented yet, I'll take it in, and bump the version by a major point on the next release.

Issue #280, opened 2016-08-02
Websocket close user


kowsari:
Hi
First of all I would like thank you for the wonderfully useful module that you have developed. This will provide alot of realtime functionality for us.

I would like to implement a presence module that allow me to keep track of exactly who is logged into the server and available to chat at any moment. 

I am a channels newbie and my question is how to get the message.user info in a websocket_disconnect. I have added the channel_session_user_from_http decorator to my consumers.py but I seem to be getting Anonymous user. For me to keep track of logged in/out users I need to be able to get the user info for both connecton and disconnection from the server.

Thanks

AlexejStukov:
Try setting ``http_user=True`` for your consumer ([doc](http://channels.readthedocs.io/en/latest/generics.html#websockets)). Then you should be able to get the user from ``message`` or ``self.message``.

kowsari:
Thanks.

I do this in my consumers.py file now

```
from channels import Group
from channels.sessions import channel_session
from channels.auth import channel_session_user_from_http
from channels.generic.websockets import JsonWebsocketConsumer

from com.utils.logger import CustomLogger


class LassoConsumer(JsonWebsocketConsumer):

    # Set to True to automatically port users from HTTP cookies
    # (you don't need channel_session_user, this implies it)
    http_user = True

    # Set to True if you want them, else leave out
    strict_ordering = True
    slight_ordering = False

    def connection_groups(self, **kwargs):
        """
        Called to return the list of groups to automatically add/remove
        this connection to/from.
        """
        return ["test"]

    def connect(self, message, **kwargs):
        """
        Perform things on connection start
        """
        CustomLogger().CORE.info('Websocket_connect. user = %s, message = %s', message.user, message)
        # transfer_user(message.http_session, message.channel_session)
        Group("notifications").add(message.reply_channel)

    def receive(self, content, **kwargs):
        """
        Called when a message is received with decoded JSON content
        """
        # Simple echo
        self.send(content)

    def disconnect(self, message, **kwargs):
        """
        Perform things on connection close
        """
        CustomLogger().CORE.info('Websocket_disconnect. user = %s, message = %s', message.user, message)
        Group("notifications").discard(message.reply_channel)

```

and this in my routing.py

`channel_routing = [
    route_class(LassoConsumer, path=r"^/")
#    "websocket.connect": "com.lassoapp.consumers.websocket_connect",
#    "websocket.keepalive": "com.lassoapp.consumers.websocket_keepalive",
[...]

PR #281, opened 2016-08-02
Easy application of decorators to Bindings

AlexejStukov:
Using the same approach as for class based consumers, so you can just set ``<decorator-name> = True`` for the class to get the decorator.

### DIFF
--- a/channels/binding/base.py
+++ b/channels/binding/base.py
@@ -6,6 +6,7 @@ from django.apps import apps
 from django.db.models.signals import post_save, post_delete
 
 from ..channel import Group
+from ..auth import channel_session, channel_session_user
 
 
 class BindingMetaclass(type):
--- a/channels/binding/base.py
+++ b/channels/binding/base.py
@@ -62,6 +62,10 @@ class Binding(object):
 
     fields = None
 
+    # Decorators
+    channel_session_user = False
+    channel_session = False
+
     @classmethod
     def register(cls):
         """
@@ -158,7 +162,23 @@ class Binding(object):
         # Run incoming action
         self.run_action(self.action, self.pk, self.data)
 
-    consumer = trigger_inbound
+    @classmethod
+    def get_handler(cls):
+        """
+        Adds decorators to trigger_inbound.
+        """
+        handler = cls.trigger_inbound
+        if cls.channel_session_user:
+            return channel_session_user(handler)
+        elif cls.channel_session:
+            return channel_session(handler)
+        else:
+            return handler
+
+    @classmethod
+    def consumer(cls, message, **kwargs):
+        handler = cls.get_handler()
+        handler(message, **kwargs)
 
     def deserialize(self, message):
         """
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Binding already has its own user code?

AlexejStukov:
I had problems getting the right user in ``has_permission``. After some searching around in the code and finding out, that i have to decorate the consumer-method, I thought it might be a good idea to go with the same approach as the consumers. Seems just easier for me.

andrewgodwin:
Well, this still doesn't handle user porting on `websocket.connect`, so I'm wary about adding it in here, that's the thing. I'd rather it was handled in the Demultiplexer, really.

AlexejStukov:
Then you would have to serialize and deserialize the user and session. Wouldn't that be a problem, if there is a custom user model?

andrewgodwin:
Err, why? I'm saying we just have people set `http_user` on the Demultiplexer, which automatically will do the session porting and set `message.user`, and then the existing code will pick it up. I don't see how this proposed method is better.

AlexejStukov:
But I would still need to decorate the ``consumer``-method of the binding with ``channel_session_user`` or ``channel_session``, since the Demultiplexer just sends the payload of the message.

andrewgodwin:
The `Demultiplexer` should absolutely annotate `message.user` if it has `http_user` set, so there is no need to add the decorator again on the binding. If it's not getting message.user that's a bug, maybe related to #280.

AlexejStukov:
``http_user`` is set and I get a message without user and session in the bindings ``consumer``/``trigger_inbound``-methods. I don't know if it's related to #280. Thought it is necessary to add the decorator to the bindings ``consumer``-function, since it's also required for normal consumer-functions like in this example from the docs:
````python
# Connected to websocket.connect
@channel_session_user_from_http
def ws_add(message):
    # Add them to the right group
    Group("chat-%s" % message.user.username[0]).add(message.reply_channel)

# Connected to websocket.receive
@channel_session_user
def ws_message(message):
    Group("chat-%s" % message.user.username[0]).send({
        "text": message['text'],
    })

# Connected to websocket.disconnect
@channel_session_user
def ws_disconnect(message):
    Group("chat-%s" % message.user.username[0]).discard(message.reply_channel)
````
Closing this for now.

andrewgodwin:
Ah! Yes, the demultiplexer currently forwards the message to a channel rather than using local routing in order to serve it, so it loses `.user` and `.session` in that process. Hrm.

andrewgodwin:
It's easy to use the routing directly instead and preserve the attributes, but it meant that workers wouldn't obey channel restrictions correctly (workers listening with a demultiplexer would service all handling channels too), but I think that's fine as it makes the whole thing more efficient anyway.

andrewgodwin:
But, then again, the message generated by the demultiplexer has different content and I'd have to special case things like `.user` in order to preserve them, which I don't like.

AlexejStukov:
I also don't think it's a good Idea to approach the relationship between ``demultiplexer`` and ``binding`` any different, than between function-based consumers like the one in the example above. Thanks again!

PR #282, opened 2016-08-02
pass subprotocols if defined

fcurella:
This PR adds a new optional setting, `WS_PROTOCOLS`, that can be passed along to Daphne.


andrewgodwin:
Hrm, I'd like at least a CHANNELS_ prefix on that setting name!

fcurella:
@andrewgodwin I've prefix the setting with `CHANNELS_`

PR #286, opened 2016-08-04
Added "callbacks" to binding

AlexejStukov:
If the user sends a message with a ``cb_id``, the binding will answer with the same ``cb_id`` and a status (and optionally other details).

### DIFF
--- a/channels/binding/base.py
+++ b/channels/binding/base.py
@@ -153,10 +153,10 @@ class Binding(object):
         self = cls()
         self.message = message
         # Deserialize message
-        self.action, self.pk, self.data = self.deserialize(self.message)
+        self.action, self.pk, self.data, self.cb_id = self.deserialize(self.message)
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Is the idea here to notify the user when an action has been completed? Surely this is done by receiving the relevant outgoing binding update - I don't really want to add something like this in.

AlexejStukov:
That's right for update and delete, but if the user creates an object, there is no way of telling if the incomming create message is caused by the users action or by somebody else, since the pk is unknown.

andrewgodwin:
True, but I still don't want to add something like this to the default bindings - this is the sort of think I would imagine being put in a third-party app as it's getting more and more opinionated.

AlexejStukov:
Ok, then I'll close this. Thanks for your time, anyway :+1:

PR #298, opened 2016-08-14
Start only one worker with the nothreading option.

ostcar:
Reuse the nothreading option of the runserver command to only start one
worker.


andrewgodwin:
Thanks!

PR #305, opened 2016-08-20
Add TransactionChannelTestCase.

furious-luke:
Use a mixin for common test-case code. This way we can have both a regular channels test-case, and a transaction test-case, too.

### DIFF
--- a/channels/tests/__init__.py
+++ b/channels/tests/__init__.py
@@ -1,2 +1,2 @@
-from .base import ChannelTestCase, Client, apply_routes  # NOQA isort:skip
+from .base import TransactionChannelTestCase, ChannelTestCase, Client, apply_routes  # NOQA isort:skip
 from .http import HttpClient  # NOQA isort:skip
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Urgh, I do hate mixins. Is there a reason you need a transaction-free channels test case? I hadn't included one initially as generally the only thing one tests with them is transactions themselves.

furious-luke:
I'm using Django's `on_commit` feature to launch a channel message. The
tests have to be run in a TransactionTestCase to work.

May I ask, for my own edification, what is it about mixins that you don't
like? It seemed like the most effective pattern to achieve the desired
result in this case.

Cheers,
Luke

andrewgodwin:
Fair. I don't like mixins because they make working out where code is coming from almost impossible, but that's more when they're exposed as a thing end-developers should use; here, the end result is still two classes, and they're in the same file as each other, so I think it's fine. I'll merge this in.

furious-luke:
Ah yes, I understand. Thanks!

Cheers,
Luke

PR #310, opened 2016-08-24
Add a link to django-cq.

furious-luke:
Hi there,

I'm working on a wrapper to help run distributed tasks using Channels as a backend. The idea is to cache the progress of the jobs in the Django DB to be able to track the progress of jobs, and to be able to know when jobs fail. It's still early work, but the basics are functional at the moment.

Cheers,
Luke


andrewgodwin:
Lovely - let me know how it goes!

PR #317, opened 2016-08-25
Update getting-started.rst to specify required Redis commands

rosco77:
Add in command to get Redis backend started along with doc links for people who may not be familiar with what Redis is or how it works.


andrewgodwin:
Sorry I missed this at the time - unfortunately you can't just run `redis-server` on some systems, and it's running by default on some others, so I'm not sure about adding it in like this. Maybe a more general mention that it should be running?

PR #322, opened 2016-08-27
Run workers in threads.

furious-luke:
Hi Andrew,

I've added a small change to allow workers to launch threads. I'm not sure this is something you actually want in the main channels repo, but thought I'd make a PR to open the discussion.

I'm currently working on `django-cq`, a background task queue for channels and I've found that launching several processes on the same machine to capitalise on multicore architectures tends to eat up a lot of memory. Running multiple threads in the same process should help to reduce the memory footprint.

This implementation was pretty quick, it could use some cleanup, but just thought I'd get your thoughts on the matter before making it nicer.

Cheers,
Luke

### DIFF
--- a/channels/generic/websockets.py
+++ b/channels/generic/websockets.py
@@ -27,6 +27,8 @@ class WebsocketConsumer(BaseConsumer):
     slight_ordering = False
     strict_ordering = False
 
+    groups = None
+
     def get_handler(self, message, **kwargs):
         """
         Pulls out the path onto an instance variable, and optionally
@@ -54,7 +56,7 @@ class WebsocketConsumer(BaseConsumer):
         Group(s) to make people join when they connect and leave when they
         disconnect. Make sure to return a list/tuple, not a string!
         """
-        return []
+        return self.groups or []
 
     def raw_connect(self, message, **kwargs):
         """
--- a/channels/management/commands/runserver.py
+++ b/channels/management/commands/runserver.py
@@ -151,5 +151,6 @@ class WorkerThread(threading.Thread):
     def run(self):
         self.logger.debug("Worker thread running")
         worker = Worker(channel_layer=self.channel_layer, signal_handlers=False)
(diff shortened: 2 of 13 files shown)


andrewgodwin:
I think the idea has merit, actually - the sweet spot for threads for a python process is 2-3 in my experience before they become a degredation.

Not a fan of some of the code layout here, though, the termed global in particular. I'd rather see an object reference passed around.

furious-luke:
> Not a fan of some of the code layout here, though, the termed global in particular. I'd rather see an object reference passed around.

Haha, yeah of course. This was proof of concept. I'll clean it up and make another commit when I get a chance. Thanks!

andrewgodwin:
No problem - thanks for looking into this. I need to take a look at `cq`, too, to make sure you have all the things I would think you'd need for a reliable task queue on channels :)

furious-luke:
Yeah by all means, please do; your input would be greatly appreciated. It's still early days, and but it works as expected for the use-cases I have. I'm not sure about performance, I suspect it could benefit from some analysis. Plus I'm nervous about potential database deadlocks as a result of the subtasks/chained tasks infrastructure, I need to think some more about how to be sure they won't happen.

furious-luke:
Hi again Andrew,

I've cleaned up the code a bit. Now uses a `WorkerGroup` class to manage the threads. One thing to note is that I've split `worker_ready` signal into `worker_ready` and `worker_process_ready`. There are occasions where certain operations need to be performed only per-process, not per-thread.

Let me know what you think.

Cheers,
Luke

furious-luke:
Hmmm, actually, I'm not sure about having moved the `worker_ready` signal into the `Worker` class. Could that be problematic?

andrewgodwin:
This is really nicely done - like the class arrangement! Happy to merge this if you're ready?

furious-luke:
Thanks! Yep, I think it should be okay.

PR #328, opened 2016-08-31
Fixed bug with _binding_group_names

AlexejStukov:
Since the old_group_names were saved as an attribute of ``instance`` in the ``pre_change_receiver`` it did not carry over to ``post_change_receiver`` and an ``AttributeError`` was raised. I think that saving ``old_group_names`` as a class attribute of the binding is the best fix for this.

### DIFF
--- a/channels/binding/base.py
+++ b/channels/binding/base.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import six
+from threading import Lock
 
 from django.apps import apps
 from django.db.models.signals import post_save, post_delete, pre_save, pre_delete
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Odd, we tested this change at the time to make sure it carried over. The method you've chosen here isn't thread-safe, so I can't merge it in as-is.

AlexejStukov:
Should be thread-save now. The failed test is a travis-ci issue.

andrewgodwin:
It's still not threadsafe as two different threads could write in the pre-change handler before both trying to read from the post-change handler; threadlocals are needed in order to pull this off correctly, and I'd still rather try and see if we could store things on the instance instead, because otherwise we rely on the two handlers always being called in order.

AlexejStukov:
Ok, after further investigation I found that the problem was in my code. So you can forget this.

PR #342, opened 2016-09-08
Added in reporting data and documentation for loadtesting

raiderrobert:


andrewgodwin:
Looks like you have the same content in the main dir and the subdir - is this intentional?

raiderrobert:
@andrewgodwin yep, it was intentional, as I thought that's what you asked for. Would you prefer the main dir make reference to the subdir thru a link instead?

andrewgodwin:
Yes please, I'd rather have a sort of index with date links up there. Sorry for not making that clearer!

raiderrobert:
@andrewgodwin done

andrewgodwin:
Thanks!

Issue #318, opened 2016-08-26
Add ability to port http_session into websocket.receive


andrewgodwin:
Since it's nice to have read/write access to the main HTTP session rather than just a `channel_session`, the session helpers should be extended to add the ability to persist the HTTP session key into the `channel_session` and have this be exposed as an attribute on the `message`.

Changes probably needed:

* New function to transfer http session key into channel session in a defined way
* Either `channel_session` or `http_session` decorator changed so it rehydrates the session from the key
* Roll calls to this stuff into the standard all-in-one decorators
* Make sure it's easily accessible from generic consumers

I would say that we should probably just make it part of the normal `channel_session` decorator, as it's going to be generally useful, and that way it'll appear wherever someone thinks it should.

(Based on #271)

PR #348, opened 2016-09-09
Update channel_session decorator to rehydrate http_session

linuxlewis:
This is my first take on a solution for #318 more feedback is welcomed.

I would like to add docs before this is merged because I think it might be a little tricky to understand. This is the use case as I understand it.

```python
@channel_session
@http_session
def ws_connect(message):
    pass

@channel_session
def ws_receive(message):
     print(message.http_session)
```

### DIFF
--- a/channels/tests/test_sessions.py
+++ b/channels/tests/test_sessions.py
@@ -76,6 +76,27 @@ class SessionTests(ChannelTestCase):
         with self.assertRaises(ValueError):
             inner(message)
 
+    def test_channel_session_http_session_key(self):
+        """
+        Tests that channel_session decorator rehydrates http_session from a stored session_key
+        """
+        message = Message({"reply_channel": "test-reply"}, None, None)
+
+        # Create a session with an http_session key
+        session = session_for_reply_channel("test-reply")
+        session[settings.SESSION_COOKIE_NAME] = session.session_key
+        session.save()
+
+        @channel_session
+        @channel_session
+        def inner(message):
+            message.http_session["num_ponies"] = -1
+
+        inner(message)
+
+        self.assertEqual(message.http_session.session_key, session.session_key)
+        self.assertEqual(message.http_session["num_ponies"], -1)
+
     def test_http_session(self):
         """
         Tests that http_session correctly extracts a session cookie.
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Yeah, this is roughly the approach I was thinking of, but it would be nice to:

* See if this works with both orderings of http_session/channel_session as decorators
* Consider making it an explicit argument to http_session if ordering silently makes it fail, and raise a loud error when it can't be done
* Add some docs

linuxlewis:
@andrewgodwin I followed your suggested and added a explicit `persist=True` (better name ideas?) option to the `http_session` decorator to handle the cases when the channel_session isn't available (missing decorator, wrong order).

I've also updated the docstrings of both functions. I considered adding a paragraph in the getting-started page section on sessions but it felt out of place in context of the tutorial. Might be worth to create an issue for a separate sessions docs page to explain further. 

If there's a better place in the docs where I could add more detail point me in the direct and I'd be happy to.

channels/tests/test_sessions.py
```diff
@@ -76,6 +76,27 @@ def inner(message):
...
+        """
+        message = Message({"reply_channel": "test-reply"}, None, None)
+
+        # Create a session with an http_session key
+        session = session_for_reply_channel("test-reply")
+        session[settings.SESSION_COOKIE_NAME] = session.session_key
+        session.save()
+
+        @channel_session
+        @channel_session
```
andrewgodwin: Why is this here twice?

andrewgodwin:
Hrm, I'd rather the `persist` option also actually controlled if it tried to persist rather than just giving a different error - it's a bit odd that right now it persists even if you pass `persist=False`. Also left a comment on one of the test cases about the doubling of decorators.

I'm starting to wonder, though, if there's a cleaner overall approach here - what you've done works fine, but my original slightly-botched API idea for this stuff is not working super great with it. Any ideas how you might change the whole session persistence API (given decorators as the application mechanism and backwards-compat concerns)?

linuxlewis:
I think the reason we're seeing it break down is because of the weird coordination between `http_session` and `channel_session` that needs to occur to make this work.  One idea for a cleaner API would be to have a separate decorator all together. It theory it would function similar to `channel_session_user_from_http`. I don't think it would conflict with any of the existing functionality either.

```python
def persistant_http_session(func):
  @http_session
  @channel_session
  def inner(message):
     #if write the session
     #else hydrate the session
  return inner
```

andrewgodwin:
Yeah, that might end up being cleaner. Perhaps `channel_and_http_session` to keep with the naming scheme?

linuxlewis:
This feels a lot cleaner. What are your thoughts?

andrewgodwin:
Yes, I like that a lot more! I'll merge it in now - if you feel like putting in a PR for the docs too, that'd be great, else I'll get to it next week.

PR #351, opened 2016-09-10
Inbound tests

Krukov:


andrewgodwin:
I actually deliberately didn't put `groups` on the base class, as I'd rather have one place for people to put the group names.

Krukov:
Ok I  revert changes around `group_names`.

andrewgodwin:
Great, thanks!

PR #352, opened 2016-09-10
Iimprovements for test client 

Krukov:
Sometimes websocket client send/receive non json-encoded/decoded messages. 
Plus it is helpfull for testing  nonwebsocket consumers by using the `HttpClient`.
Example:  we have group of related consumers (some of them for websocket channels and other are internal) and that is onerously to create a `Client` for a "single" call.

### DIFF
--- a/channels/auth.py
+++ b/channels/auth.py
@@ -1,4 +1,4 @@
-import functools
+    import functools
 
 from django.contrib import auth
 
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Hrm, maybe we should have this as a `json` keyword argument rather than automatic? I'm worried that having it automatic might lead to people being confused if they pass something in that's valid JSON on a channel receive and it comes out as not a string (e.g. the string `12345` would turn into an int suddenly)

andrewgodwin:
Should `json` default to `True`? That feels backwards incompatible.

Krukov:
Yes. At the previous version of  'receive` we decoded message text all the time at `HttpClient`. So using `json=True` doesn't feel backwards incompatible. Or I miss something?

andrewgodwin:
Oh, yes, you are right. It's fine as is then.

PR #358, opened 2016-09-12
Remove 'onopen' for compatibility and usability

Bobby-Miller:
socket.onopen doesn't work as expected for me on chromium based browsers (chrome and vivaldi). Instead, I recommend suggesting a straight send command in the tutorial. In addition to being more compatible, this also is more directly actionable (user can try .send over and over again, opening browsers of different types, etc.) A small change, but this is what I ended up doing, and I was stuck for a few with the .onopen attempts.

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -138,9 +138,8 @@ need to change the socket address if you're using a development VM or similar)::
     socket.onmessage = function(e) {
         alert(e.data);
     }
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Onopen works as expected for me in Chrome, and is part of the RFC spec for WebSockets, so I'm not going to remove this without proof that it's necessary.

PR #360, opened 2016-09-12
JavaScript console example update

navyad:
Calling defined function ``onopen`` in JavaScript example. when only called then alert for message will come.

selection_005


andrewgodwin:
Why would you do this? It will call `onopen` even if the connection is not established - I don't think this makes sense.

PR #361, opened 2016-09-12
reply and response are confusing

daonb:
Hi,
Started reading the docs and found them a bit confusing. Based on the attribute name, `response` belong to the HTTP domain and `reply` to the channel domain.

Hope it helps,

Benny


andrewgodwin:
Thanks, this is probably left over from when they *were* called response channels.

PR #370, opened 2016-09-16
Typecasted server port from integer to string.

ayushjain19:
Converted self.message['server'][1] form integer to string using str() function so that SERVER_PORT can be received as a string in HttpRequest.META


txels:
Needs fixing the corresponding test case:

```
  File "/home/travis/build/django/channels/channels/tests/test_request.py", line 66, in test_extended
    self.assertEqual(request.META["SERVER_PORT"], 80)
AssertionError: '80' != 80
```

andrewgodwin:
I'd also like to be sure that `str` is the right thing to use here, rather than `six.text_type` or `six.binary_type`.

PR #371, opened 2016-09-16
string should be compared with string

ayushjain19:
request.META["SERVER_PORT"] gives string, and hence should be compared with a string in assertEqual() method. Therefore, I have converted integer value 80 to string "80".


andrewgodwin:
Can you combine this with the other PR please? I can't merge unless the tests are passing and either of these will pass if not combined.

PR #375, opened 2016-09-22
Add conventional request.META['PATH_INFO']

benjaoming:
As per discussion on #django-channels

We claim it's conventional because of this issue: benjaoming/django-nyt#27

Maybe it's also documented somewhere... that would only make it more justified :)


andrewgodwin:
Thanks - not sure how I missed that one.

PR #378, opened 2016-09-23
Fix type for request.META['SERVER_PORT']

### LINKED ISSUE #366: request.META['SERVER_PORT'] should be a string (?)
Found while debugging an issue with Opbeat: https://github.com/opbeat/opbeat_python/issues/131

ASGI request in django channels initialises `request.META['SERVER_PORT']` using an integer value: https://github.com/django/channels/blob/master/channels/handler.py#L68

```
ipdb> self.message['server']
['127.0.0.1', 7000]
```

This later results in:
```
ipdb> request.META['SERVER_PORT']
7000
```

This appears to be in contradiction with the django docs (https://docs.djangoproject.com/en/1.10/ref/request-response/#django.http.HttpRequest.META) that state:

> SERVER_PORT – The port of the server (as a string).

Which causes the mismatch and failure to validate when posting to opbeat.

To me it feels that port should be handled as an integer rather than a string, but I see that using a string goes back to the WSGI spec https://www.python.org/dev/peps/pep-0333/#environ-variables and probably is a heritage from CGI times.

I wonder which is the correct way forward. If this should be a string in ASGI requests as well, I'm happy to volunteer a PR.

txels:
Django documentation states that it is a string
(https://docs.djangoproject.com/en/1.10/ref/request-response/#django.http.HttpRequest.META):

> SERVER_PORT – The port of the server (as a string).

Fixes #366 
Recreates incomplete #370

### DIFF
--- a/channels/handler.py
+++ b/channels/handler.py
@@ -66,7 +66,7 @@ class AsgiRequest(http.HttpRequest):
             self.META['REMOTE_PORT'] = self.message['client'][1]
         if self.message.get('server', None):
             self.META['SERVER_NAME'] = self.message['server'][0]
-            self.META['SERVER_PORT'] = self.message['server'][1]
+            self.META['SERVER_PORT'] = six.text_type(self.message['server'][1])
         # Handle old style-headers for a transition period
         if "headers" in self.message and isinstance(self.message['headers'], dict):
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Thanks!

PR #382, opened 2016-10-01
Add GitHub issue template

ssteinerx:
Add an issue template to ask the questions that have to be asked on every bug report etc.

Basic version information, behaviour, expected behaviour, how are you connected etc. that seems to have to be established for every issue.

Should save some time...

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,26 @@
+## Submit a feature request or bug report
+
+### Versions
+
+OS/Version:
+
+Django:
+
+Channels:
+
+Daphne:
+
+Twisted:
+
+### Connection Information
+
+**Are you connecting to daphne/runserver directly? If not, please describe the connection.**
+
+**What is the current behavior? (please include logs/output)**
+
+
+**What is the expected or desired behavior?**
+
+
+**Any other/relevant information:**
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Thanks - I had forgot you could do this.

shaib:
We're now seeing a limitation of github issues: The issue describing this very change would not fit the template very well.

ssteinerx:
The template is just a suggestion and asks for the information common to issues related to the project.  It's not perfect, but should reduce the repeated "What version of channels..." that seem to have to be asked on almost every issue related to channels itself.

shaib:
Don't get me wrong: I'm sure this template is an improvement to the project, and it's a Good Thing(TM) that you added it. The criticism wasn't against this patch, but against GitHub Issues.

senden9:
Comments are also possible in templates. Syntax:
`<!--- This comment is only visible in the write mode. Not in the final issue. -->`
We could add a comment that says that this is just a suggestion if this makes the things more clearly.

andrewgodwin:
I will modify the template to be a bit more generic regardless.

ssteinerx:
Wasn't meant to be final, more of a reminder...

andrewgodwin:
I appreciate the reminder! Thanks for bringing this up.

PR #387, opened 2016-10-07
added http_timeout as a command line option for runserver

RockHoward:
Only added http_timeout. I am prepared to add other command line options upon request.


RockHoward:
Those tests failed because the test expected http_timeout to be passed in to the server as None instead of 60.

This seems bogus since the value was hard-coded to 60 before I made my change.

andrewgodwin:
It's just mocking so it's very fragile - just fix the assertion in the pull request too so it passes.

RockHoward:
Gotcha. Hopefully I will have a few minutes to get this done this evening.

andrewgodwin:
Lovely, thanks!

PR #396, opened 2016-10-12
Fix for streaming responses being cached in memory.

furious-luke:
Found a bug whereby streaming responses would try to be cached entirely in memory. Was causing views that stream a lot of data to timeout.


andrewgodwin:
Ah, nice catch, and before it's even in a stable release.

Issue #115, opened 2016-04-04
Delay protocol server


andrewgodwin:
A "protocol" server (more just a special process) that listens on a given channel for messages with `channel`, `content` and `delay` keys, and re-inserts them onto the destination channel after the specified delay. Used to have tasks happen after a delay without clogging up the main channel layer with them until they're ready to be worked on.

The suggested channel name is `channels.delay`, with the idea that the `channels` namespace can grow to cover other generic (non-protocol-specific) things like this.

This might also share an implementation with #74.

linuxlewis:
I think this can be closed now

hoaivan:
This is wonderful. Any schedule to make this into a release 0.17.4?

andrewgodwin:
It'll be in the 1.0 release when I can get it out (this likely involves my wrist healing up first!)

Issue #74, opened 2016-02-26
Running scheduled jobs


erikash:
Hi, 
Kudos for the great job! 
This will be definitely take django to the next level! 

Are you planning to include support for scheduled jobs? Like celery beat? 

Any suggested workarounds? 

Best, 
Erik.

andrewgodwin:
Yes, the intention is to have some sort of command that can push messages onto channels periodically. In the meantime, I suggest writing a custom management command that does this and calling it using `cron`.

erikash:
Thanks, sounds great

avayo:
Hello Andrew. Thank you for the great project!
Can you (or someone else) suggest the best way to send signals to channels via cron?

andrewgodwin:
For now, I recommend writing a management command that sends the signals onto the channels, and calling that from cron.

mrasband:
For anyone else coming here and not being sure where to start (like me), a quick example:

```python

# settings.py
CHANNEL_LAYERS = {
    # This is the channel layer id, you can have a bunch if you want
    'default': {
        'BACKEND': 'asgi_redis.RedisChannelLayer',
        'CONFIG': {
            'hosts': [os.environ.get('REDIS_URL', 'redis://localhost:6379')],
        },
        # this defines the channels on the layer (in tutorials you probably see websocket.connect)
        'ROUTING': 'app.routing.channel_routing',  
    }
}

# app/routing.py
channel_routing = {
    # You can name your channel anything you want
    'your_channel_name': lambda x: print('Channel Triggered Event!'),
}

# ./manage.py shell
from channels import Channel, channel_layers

# default here matches your layer name (from settings.py) that can delegate to your_channel_name
c = Channel('your_channel_name', channel_layer=channel_layers['default'])
c.send({})
```

andrewgodwin:
Yup, that's a great example - thanks @mrasband! I'll see if I can work stuff for calling from cron into the docs - it also looks like we're going to have some work start on a scheduled task interface server soon (similar to celerybeat), which might be better if you want things very frequently or can't set cron tasks.

mrasband:
Thank you @andrewgodwin for the awesome work, this has helped swing me back to using Django more :)

rajasimon:
@andrewgodwin you mention "scheduled task interface" server. Is developed fully...! Shall I try now?

andrewgodwin:
I believe @jacobkm was looking into it, but I don't know what sort of progress he's made, or if he's found the time to start yet.

[...]

PR #401, opened 2016-10-17
Delay Protocol Server

linuxlewis:
Here's a first version of an implementation for #115.

Welcome any feedback or suggestions, here's a few questions I am looking for thoughts on:
- How/should delay messages be persisted?
  - I think a django model is an option here. Celery beat uses shelves by default which could also be done.
- Should the delay channel name be configurable?
## Quick start

Here's how you run the server.

``` bash
python manage.py rundelay
```

Then send a message to `channels.delay` that has the keys `content`, `channel`, and `delay` to delay it.

``` python
content = { 'content': {'x': 1}, 'channel': 'test.math', delay: 10 }
Channel('channels.delay').send(content, True)
```


andrewgodwin:
I think a Django model is probably the best way, given that Channels is a Django app. I'd be tempted to say maybe the delay server should even be its own separate Django app so the models can be limited to install if you add it to INSTALLED_APPS?

I think the channel name should not be configurable - the protocol specs so far all have fixed names (e.g. `http.request`) as that's part of what you code against. I do wonder if we should call it `__channels__.delay` or something more similar though, or if the `channels` namespace could just be kept as a reserve.

linuxlewis:
I think keeping the `channels.*` namespace reserved is reasonable. I might suggest to go as far as to enforce this in the router or `runworker` with an error so it's clear to users.

I actually think a separate django app is a great idea. Especially if this continues to evolve into #74

How do you feel about making the storage configurable? I think a django model is a fine default but maybe it's possible someone would prefer a local memory or redis backend option.

andrewgodwin:
I would say no to configurable storage (my policy is to say no to generic abstractions unless there's a real need) - if someone wants different behaviour, I think they should write their own delay server against the specification we'll come up with, since that's the real abstraction (and they might be able to structure the storage interaction better, e.g. async calls to the datastore if possible)

linuxlewis:
Added some commits here. Created the channels.delay app and moved everything over. Messages are also now stored using a django model.

I added support for an `interval` parameter so messages can be dispatched periodically.

Any feedback is welcomed

andrewgodwin:
I think it looks pretty good - some questions:

* Do you think it's worth refactoring the Worker code to share between this and `runworker`? I've been pondering having a basic worker loop in `asgiref` which might help this too.
* Could you add a spec for how delays work? There's sub-specs for ASGI in the docs already, which also makes me wonder if we should call it `asgi.delay` or something else without `channels` in it.
* Given that it would be nice to also have crontab-like periodic events emitted from the server without having to initiate them first (as per the "delay every X seconds" stuff you have), maybe it would be best to remove the delay-every-X and have a separate python/statically-configurable method for sending periodic events (the implementation doesn't have to make it in, just want to know what you think)

linuxlewis:
- I think this would be nice to have. Since there was a lot of crossover with `runworker`, I considered actually modifying it to handle the delay logic and exposing it as a  `--delay` option. With that said it feels like out of scope of this PR but definitely a quick refactor.
- Yes defintely! When you say sub-specs I'm assuming you mean something like this? (http://channels.readthedocs.io/en/stable/asgi/email.html)
- I like this idea a lot and it's something I thought about but didn't have a clear idea on how It would work. The current interface accepts a delay parameter as the number of seconds to delay. Do you see this as augmenting that interface to pass an importable string? `{'channel': 'test', 'delay': 'module.callable_that_returns_seconds', 'content': {'x': 1}}`. Or do you see this callable being executed on every worker run?

andrewgodwin:
* Let's keep that out of this PR but consider it once it's merged, as I think it would be widely valuable.
* I do mean something exactly like that.
* I mean not having it configured via channels at all (as you can lose messages over channels), but having just a Django setting or similar that configures it. Just a list of channel, payload, and interval would work.

linuxlewis:
Ok put some more commits up.

- There's now an official ASGI spec. I changed the name of the channel from `channels.delay` to `asgi.delay` to make it less of a channels specific thing.
- interval functionality has been removed
- changed default unit from seconds to milliseconds. This is largely from feedback other developers I spoke to about the feature.

Let me know your thoughts.

andrewgodwin:
Yup, this seems lovely now. Sorry about the delay in getting back to you - anything else you want to add or should I merge it?

linuxlewis:
No worries, appreciate your feedback on this. Merge away 👍

andrewgodwin:
Thanks for your work on this!

sachinrekhi:
Great work on this guys! Looking forward to giving it a try.

One question: after looking at the implementation, it seems like if I ever need to reschedule / remove some delayed tasks, I can directly manipulate the Django model to change the due_date or just delete the model object entirely correct? It looks like in the delay server run loop it makes a live db query each time, so looks like I shouldn't run into any caching issues or such, right?

andrewgodwin:
Well, that's not an officially documented way, so I can't say it'll work forever, but yes, that would work for now.

Issue #383, opened 2016-10-01
API proposal: consumer class demultiplexing


raphael-boucher:
# Current situation on demultiplexing

At the moment, we can't use consumer classes with the websocket demultiplexer.

If actions needs to be performed on connection or disconnection, it has to happen in the demultiplexer.
It forces some coupling between the channel and the demultiplexer, does not give reusability and goes beyond what a demultiplexer should do (transfer messages to the specified channel).

``` python
# Example of a demultiplexed channel for a presence use case
#
# Demultiplexer and consumer are coupled, connection and disconnection happen in the demultiplexer, 
# the rest in functions. Group management happens in the demultiplexer, which should not be the case.

class Demultiplexer(WebsocketDemultiplexer):
    mapping = {
        "presence": "presence.receive",
    }
    http_user = True
    slight_ordering = True
    groups = ['presence-all']

    def connect(self, message, **kwargs):
        """Called when a WebSocket connection is opened."""

        # This is related to the presence consumer and can not be reused easily
        self.group_send('presence-all', {
            message.user.id: 'welcome a new user'
        })
        message.user.online = True
        message.user.save()

    def disconnect(self, message, **kwargs):
        """Inform other users of the disconnection"""

        # This is related to the presence consumer and can not be reused easily
        self.group_send('presence-all', {
            message.user.id: 'Bye by to a user'
        })
        message.user.online = False
        message.user.save()


@channel_session_user
def presence_receive(message):
    # simple echo
    Demultiplexer.send('presence', message)


# Add routing for both classes
channel_routing = [
    route_class(Demultiplexer, path=r'^/ws/$'),
    route('presence.receive', presence_receive),
]
```
# New proposed API: the class consumer demultiplexer

I offer to implement a websocket demultiplexer that handle classe consumers, based on the `WebsocketDemultiplexer`. This demultiplexer would forward all messages, including connection and disconnection to the channels of all given consumer classes.

This would work with a new consumer class, called `DemultiplexedConsumer` based on the JsonWebsocketConsumer. 
A new attribute `stream_name` gives the name of the demultiplexed stream. This name would be used to generate dynamically the `method_mapping`:

``` python
# Default mapping for a websocket class
method_mapping = {
    "websocket.connect": "raw_connect",
    "websocket.receive": "raw_receive",
    "websocket.disconnect": "raw_disconnect",
}

# Dynamic mapping for the demultiplexed consumer class
method_mapping = {
[...]

PR #420, opened 2016-11-01
Add demultiplexer for class-based consumers (#383)

raphael-boucher:
Avoid coupling between the demultiplexer and consumers.

Explanations are in https://github.com/django/channels/issues/383 .

In short, I chose to forward a multiplexer object to the consumer. This object can be used to send messages through the demultiplexer without depending on the stream used.

### DIFF
--- a/channels/binding/websockets.py
+++ b/channels/binding/websockets.py
@@ -3,7 +3,7 @@ import json
 from django.core import serializers
 from django.core.serializers.json import DjangoJSONEncoder
 
-from ..generic.websockets import WebsocketDemultiplexer
+from ..generic.websockets import WebsocketMultiplexer
 from ..sessions import enforce_ordering
 from .base import Binding
 
@@ -40,7 +40,7 @@ class WebsocketBinding(Binding):
     # Outbound
     @classmethod
     def encode(cls, stream, payload):
-        return WebsocketDemultiplexer.encode(stream, payload)
+        return WebsocketMultiplexer.encode(stream, payload)
 
     def serialize(self, instance, action):
         payload = {
--- a/channels/exceptions.py
+++ b/channels/exceptions.py
@@ -33,7 +33,14 @@ class RequestAborted(Exception):
 
 class DenyConnection(Exception):
     """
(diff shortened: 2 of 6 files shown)


andrewgodwin:
Some thoughts:

* It's worth removing the old demultiplexer and docs with this patch, I don't want to have two!
* I quite like the design, but it still results in `self.send` doing the wrong thing and not raising an error in a multiplexed consumer. Do you have any ideas how we could at least make it error?

raphael-boucher:
If the reply channel is set to None, I can't use it to send a reply back. It could be done by cloning the message, but it feels quite messy, and it might cause issues on lower stack layers.
On the other hand, if you send through the consumer without using the multiplexer, it's not incoherent to send a normal message. 

I started working on removing the old demultiplexer, but I will need some help to avoid breaking all the binding features depending on it.

andrewgodwin:
Well, I just want to break anything that would silently work but do the wrong thing (like `self.send`). Part of me thinks sending should just be totally wrapped up so it can be intercepted somehow, but I'm not sure of the best approach there.

I'm happy to help remove the old stuff once this lands in a way where there's minimal breakage; I expect some, but the whole API is marked as alpha, so there's a bit of a pass.

raphael-boucher:
Until it's possible to hook on `send` (I would like to avoid the scope of this PR becoming too big), would it be ok to throw an error `self.send` explaining that the demultiplexer should be used?
I didn't want to do that at first because `self.send` is used in the demultiplexer, but I could simply use the `reply_channel` in the multiplexer and have it work.

andrewgodwin:
Yes, that's fine by me. As long as people can't shoot themselves in the foot without an explicit error happening.

raphael-boucher:
Data binding test are failing, if you see how to adapt them for the class demultiplexer it would be very helpful, since I don't know this part very much.

andrewgodwin:
I'll take a look at some point in the next few weeks, my wrist is healing right now so it's hard to program!

andrewgodwin:
I have manually merged this and fixed the interaction with databinding!

raphael-boucher:
Awesome!

Issue #356, opened 2016-09-11
Forwarding connect message to another channel and session creation


raphael-boucher:
I'm forwarding the connection message from a channel that creates the session to another channel that suppose the session exists, like this:

``` python
class AnotherConsumer(BaseConsumer):
    # get the session, that needs to be already created
    channel_user_session = True

    method_mapping = {
        'another.connect': 'raw_connect',
    }

class MyConsumer(JSONWebsocketConsumer):
    # create the session from http cookies sent on connect message
    http_user = True

    def connect(message):
         Channel('another').send(message.content)
```

When using `runserver`, the session is lost in the second consumer most of the time (it works randomly and rarely).
When using `runserver --no_worker` + `runworker` or setting a breaking point just before the fowarding, it works fine all the time.

A solution I have right now is to create the session again on the second consumer, but I would like to avoid that if possible.
I suppose the session is not fully registered when the message is forwarded to the second channel in `runserver` mode. Any idea how to ensure that ?

andrewgodwin:
Can you say what you mean by "the session is lost"? I suspect you're encountering an ordering problem (see http://channels.readthedocs.io/en/latest/getting-started.html#enforcing-ordering) and the other consumer is being called before your one exits, so the session is not yet persisted.

I'm not sure quite how you would fix this, given that you cannot use `enforce_ordering`. Maybe we need a decorator that delays sends till function completion (and the session is written), or maybe we should write the session before connect is called.

raphael-boucher:
It feels like the session is lost since it exists in the first consumer, but it was a poor formulation.
The session is actually not yet created, for the reason you just said:

```
#in channel_session

  # Run the consumer
        try:
            return func(message, *args, **kwargs)
        finally:
            # Persist session if needed
            if session.modified:
                session.save()
```

Is there a reason to save the session only after connect is called?

andrewgodwin:
Well, it needs to be saved at the end in case it's been changed, but we could save it before as well.

raphael-boucher:
Fair enough!

eternaldensity:
I encountered this problem too.
Basically, while the transfer_user function executes before the connect consumer, the session.save doesn't execute until after the connect consumer.  So it's not guaranteed to have saved the transfer before any other consumers (that are sent messages from the connect consumer) get executed (or more specifically, before channel_session invokes for them.)
[...]

PR #421, opened 2016-11-01
Save session before invoking consumers (#356)

raphael-boucher:
With demultiplexers, consumers could be invoked after the session has been created but not saved. These consumers wouldn't find the session in the database.

See also #356

### DIFF
--- a/channels/sessions.py
+++ b/channels/sessions.py
@@ -61,6 +61,7 @@ def channel_session(func):
         message.channel_session = session
         # Run the consumer
         try:
+            session.save()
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Do you have a unit test that can prove this was broken before? My reading of the creation portion is that it must save during creation to make sure it exists, thus there's no need to save it again.

andrewgodwin:
Closing this in favour of the approach in #486, as that would also fix it but result in less saves if nothing changes.

PR #424, opened 2016-11-05
Updating Channels status regarding inclusion into django core

jibaku:
Updated docs to add information related to Channels being a Django official project.


jibaku:
And maybe it makes sense to remove the docs/cross-compat.rst file for the moment ?

andrewgodwin:
I agree cross-compat should probably just go. Do you want to include that into this PR?

jibaku:
@andrewgodwin Yes. I've removed the file/reference to it and updated the PR.

andrewgodwin:
Thanks!

Issue #414, opened 2016-10-26
Enable server specify code and reason when closing a websocket connection


tyler46:
Hi, 

It would be nice to have a way to specify the close code or close reason when server decides to close a web socket connection.

(Issue was first mentioned as #327)

rixx:
I'm working on this at DUTH sprints right now. ✨

We decided to go with the closing code, but not the optional additional reason.

tyler46:
Hi all, many thanks for the effort. I will try to implement the the optional additional reason and come back to you!
:+1:

andrewgodwin:
Closing this as we indeed added in closing code, and I'm fine leaving reason out.

PR #426, opened 2016-11-05
Add closing response codes

rixx:
Added both to spec and implementation. Regards #414.

### DIFF
--- a/channels/generic/websockets.py
+++ b/channels/generic/websockets.py
@@ -95,7 +95,7 @@ class WebsocketConsumer(BaseConsumer):
         """
         message = {}
         if close:
-            message["close"] = True
+            message["close"] = close
         if text is not None:
             message["text"] = text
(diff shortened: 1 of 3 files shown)


bastbnl:
I've got a question concerning this PR. I'm using channels==0.17.3 with daphne==0.15.0 on Django==1.10.3 and I'm closing the connection from a consumer. The connection is closed using code 1005 when `close=True` (verified from angular-websocket on Chrome 54.0.2840.99 m):

`message.reply_channel.send({"close": True})` 

Am I doing something wrong?

rixx:
This has been merged in django/Daphne#58, which has been a while after the 0.15.0 release. You'll have to either use a non-released Daphne version from Daphne's master or wait for @andrewgodwin to release the next Daphne version.

bastbnl:
Thanks! Will wait for it

PR #429, opened 2016-11-07
added instructions in order to run daphne with django-configurations

fpischedda:
added instructions in order to run daphne with django-configurations defined settings

### DIFF
--- a/docs/deploying.rst
+++ b/docs/deploying.rst
@@ -164,6 +164,20 @@ Then, you can run Daphne and supply the channel layer as the argument::
 
     daphne my_project.asgi:channel_layer
 
+If you are using django-configurations your asgy.py may need some more tweaks,
+here is an example::
+
+    import os
+    from channels.asgi import get_channel_layer
+    from configurations import importer  # <- configurations settings loader
+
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Given that django-configurations is not a core part of Django, I'm not sure this belongs in the main docs. Is there a specific side-effect you could highlight instead that causes the problem?

fpischedda:
Hi,

you are right that this notes on the documetation are more suited for django-configurations project; I wrote this PR because I was deploying a web app developed by my team and having seen django-channels for the first time I've headed to channels' deploy documentation, followed the instructions to create a asgi.py module, ran daphne and got a nice ImproperlyConfiguredError while the "django" project was running perfectly; after some tweaking I have found out the problem, fixed it and wanted to share.

So you are right that this has to be documented to the django-configurations side but it also true that a first timer with channels will go to read its deployment documentation and probably follow my very same path getting stuck like me :)

That said it is ok to reject my pull request but it would be nice to add some notes about configurations like DRF does in its documentation, mentioning useful third party apps integration.

andrewgodwin:
Thanks for understanding. Is there a particular thing that you think would be good to include in the main docs, like how to generically avoid ImproperlyConfiguredError?

PR #434, opened 2016-11-14
build endpoint description strings from runserver arguments before ca…

mcallistersean:
This pull request switches the runserver command of channels to using twisted's endpoint description strings internally instead of the usual host/port arguments passed directly to the daphne server class.
This is related to https://github.com/django/daphne/pull/37 which would need to land before this can be safely merged.


andrewgodwin:
I'll manually bring this one down and run the tests against master daphne to merge it, so it might take a bit.

andrewgodwin:
Got this merged in eventually now I'm in the 1.0 release process. Sorry it took so long!

channels/management/commands/runserver.py
```diff
@@ -2,7 +2,7 @@
 import sys
 import threading
 
-from daphne.server import Server
+from daphne.server import Server, build_endpoint_description_strings
```
mguerra502, line 5: That import is giving me problems when I try to run:

python manage.py runserver

I get the following error:

ImportError: cannot import name 'build_endpoint_description_strings'

Do you happen to know what could be the problem for this?

PR #435, opened 2016-11-17
runserver should respect FORCE_SCRIPT_NAME setting

jan-lugfl:
I am currently playing around with channels 0.17.3 in an old Django 1.8 ( I know its old but have to stick with i for now due to MSSQL ).

I have nginx running in the frontend as a proxy to path based seperate different projects from each other on the same hostname, which works without any problem with runserver or WSGI without channels activated. But as soon I enable channels the build-in runserver does not respect the FORCE_SCRIPT_NAME setting so the paths in rendered templates are wrong as the prefix set via FORCE_SCRIPT_NAME is missing.

As daphne already seem to have support for root_path I just pass FORCE_SCRIPT_NAME as root_path which fixes the issue.
I don't know if a new settings value would be more appropriate though.
In addition I stumbled over an issue that resulted in wrong path setting as the the PATH does not contain the root_path, so I changed so it only gets stripped when present.

### DIFF
--- a/channels/handler.py
+++ b/channels/handler.py
@@ -44,7 +44,7 @@ class AsgiRequest(http.HttpRequest):
         # Path info
         self.path = self.message['path']
         self.script_name = self.message.get('root_path', '')
-        if self.script_name:
+        if self.script_name and self.path.startswith(self.script_name):
             # TODO: Better is-prefix checking, slash handling?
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Yup, this seems like a smart move. Thanks!

PR #437, opened 2016-11-26
Updated channels loadtesting results

raiderrobert:
@andrewgodwin, finally pulled together the changes. Nothing significant really. Mostly just some rewording and clarifications. I could put in a lot more words, but I spent a lot of time this afternoon carefully considering what not to say.

My one thought of what I could include, but wasn't sure was (1) an executive summary and (2) suggestions for possible ways of expanding on this report later.


andrewgodwin:
I think a summary could be nice, as long as it was balanced and conservative in its conclusions.

PR #448, opened 2016-12-07
 Use save's update_fields in serialize_data

AlexejStukov:
pass save's kwargs to serializes so it can access `update_fields` and other.

### DIFF
--- a/channels/binding/base.py
+++ b/channels/binding/base.py
@@ -70,6 +70,9 @@ class Binding(object):
     # Decorators
     channel_session_user = True
     channel_session = False
+    
+    # the kwargs the triggering signal (e.g. post_save) was emitted with
+    signal_kwargs = None
 
     @classmethod
     def register(cls):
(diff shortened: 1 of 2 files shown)


AlexejStukov:
With this serialize and serialize_data can react on `update_fields`.

andrewgodwin:
I'm not sure I like the way this passes implementation details of Django signals all the way through to the databinding serializer, or how it only returns the updated fields rather than the whole object every time (I like consistency, and consistency is returning the whole object IMO).

What's the drive to add this? Is there an alternate way that doesn't pass down all the signal arguments and require changing the signature of `serialize`, maybe by special-casing updated fields and using `self` so people could opt into this if they wanted it?

AlexejStukov:
You are probably right about passing the kwargs. My drive for that was, that I periodically updated only one field of a model and did not want to publish the outdated other fields. To ensure that they are always up to date I would have to fetch the instance every time. Are there even other use cases for `update_fields`?

andrewgodwin:
I think that would be the only case; I would prefer to just populate `self.signal_kwargs` in the main repo, not change serialize at all, and let people do update logic like yours in their own subclass.

AlexejStukov:
OK, I changed it accordingly.

AlexejStukov:
Thanks!

Issue #452, opened 2016-12-13
Test HTTP client method may generate invalid cookies `HttpClient.set_cookie`


rectangletangle:
When writing unit-tests around the test HTTP client, my cookies seem to be getting mangled when adding an additional field.

Example:
```
from channels.tests import ChannelTestCase, HttpClient

class MyTests(ChannelTestCase):
    def test(self):
        httpclient = HttpClient()
        httpclient.set_cookie('foo', 'bar')
        ...
```

The culprit appears to be this routine, where it joins the cookie fields on `&`.
https://github.com/django/channels/blob/f4f45dbb9f63c4085d65331dd534740978ab6623/channels/tests/http.py#L126

I discovered this while trying to parse the test client's cookies with Django's parsing routine. 
``` 
from django.http.cookie import parse_cookie
```

I believe replacing the function with this should do the trick (note the join on `'; '`):
```
def _encoded_cookies(cookies):
    """Encode dict of cookies to ascii string"""
    return ('; '.join('{0}={1}'.format(k, v) for k, v in cookies.items())).encode("ascii")
```

Anyone know if there's a cookie serialization routine in Django core, or Python's stdlib? 

I can implement a PR, if necessary.

Krukov:
You right. I don't know why I used `&` :suspect:  and I could not find a simple method to do a cookie serialization by 'standart' way. Anyway I sure that PR is welcome

rectangletangle:
Reminds me of URL params or something lol. Great work on channels BTW!

I'll try to get a fix PR up soon.

Krukov:
Let's close it.

PR #453, opened 2016-12-14
Valid cookie serialization for the test HTTPClient

rectangletangle:
https://github.com/django/channels/issues/452

This PR attempts to fix the `HTTPClient.set_cookie` method, which was serializing cookies into an invalid format. 

It may be worthwhile to re-implement the function using the stdlib:
https://docs.python.org/3/library/http.cookies.html

Any suggestions on where to put tests for test utilities?

### DIFF
--- a/channels/tests/http.py
+++ b/channels/tests/http.py
@@ -6,6 +6,7 @@ from django.apps import apps
 from django.conf import settings
 
 from django.utils.six.moves import http_cookies
+from django.http.cookie import SimpleCookie
 
 from ..sessions import session_for_reply_channel
 from .base import Client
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I would put them in the same tests directory in a new file, just make a test that tests the utility. Do you want to do that before I pull this in?

rectangletangle:
Yep I'll add a test, it's pretty easy to prevent regressions here.

rectangletangle:
Added some tests, ready for review.

rectangletangle:
Looks like it's failing with certain Python versions, will look into a fix

rectangletangle:
Added some fixes, should be ready to merge (assuming it passes review).

andrewgodwin:
Lovely, thanks!

PR #467, opened 2017-01-09
Accept Connection at WebsocketConsumer

Krukov:
Added  sending "accept" message at `WebsocketConsumer` as default behavior for `connect` method. I wanted to add it at `raw_connect` but it had break `Demultiplexer` logic.

Also I think that we must pass `message` itself to the `trigger_inbound`, because we need to have access to the `user` or to the `reply_channel`


andrewgodwin:
Yup, good call.

PR #468, opened 2017-01-09
User friendly way to close websocket with code (via CloseException)

Krukov:
@andrewgodwin what do you think about it?

### DIFF
--- a/channels/exceptions.py
+++ b/channels/exceptions.py
@@ -1,3 +1,7 @@
+from __future__ import unicode_literals
+import six
+
+
 class ConsumeLater(Exception):
     """
     Exception that says that the current message should be re-queued back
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Hmm, I avoided doing this so far as I didn't want another way to do it, but there are definitely advantages, not least allowing some of the decorators to close stuff properly.

My main concern is how it's websocket specific, but has to be written directly into the worker; it would be nice if we could have a slightly more generic way that allowed closing of other protocols like this without people having to have their own Worker class. I'm thinking something like a base ChannelSocketException where subclasses get `.run()` or something when they're caught by the worker, and so the WebsocketCloseException can have a `run()` that sends `close=code`

Krukov:
:+1:  I will correct it soon.

Krukov:
@andrewgodwin It is more suitable solution?

andrewgodwin:
Yes! If you could fix the import sorting error I'll merge this.

Krukov:
Done!

andrewgodwin:
Thanks!

Issue #377, opened 2016-09-22
Validation error path for data binding


andrewgodwin:
There needs to be a standard way to return errors for inbound data binding back to the client.

andrewgodwin:
Data binding is no longer in Channels as of v2, so closing.

PR #469, opened 2017-01-09
Form base validation for inbound bindings

Krukov:
Regarding #377 .
I didn't find any standard way to inform client about errors, So I suggest to use django forms and conventional format for json messages.


andrewgodwin:
I'm not sure I would want to include form usage with a standard way about errors - maybe just having a standard way and a hook to send them would be enough? Also, this includes the disconnection exception that is also in #468.

Krukov:
Ok. I'll make new PR.

PR #473, opened 2017-01-11
more consistent metrics

vartec:
More consistent metrics, in particular for consumers such as Graphite, CloudWatch etc. you don't want aggregated numbers per second, you want number since previous call. Since we don't want to track all the clients, total since start is the next best thing, because it can be easily calculated by the metrics consumer `count_since_last = pervious_count - current_count`

### DIFF
--- a/docs/asgi.rst
+++ b/docs/asgi.rst
@@ -367,17 +367,20 @@ A channel layer implementing the ``groups`` extension must also provide:
 
 A channel layer implementing the ``statistics`` extension must also provide:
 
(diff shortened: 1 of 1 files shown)


andrewgodwin:
`queue_depth` doesn't match the rest of the names. Maybe call it `messages_pending`?

vartec:
@andrewgodwin fair point. How about `messages_count`? Should it stay as is, or be `messages_processes`? (`messages_processed_count` would be a tad to explicit)

andrewgodwin:
Lovely, that seems a lot more consistent.

PR #480, opened 2017-01-11
Send all messages before ChannelTestCase.get_next_message looks for a message

ostcar:
In the documentation there is an test example, which does currently not work:
http://channels.readthedocs.io/en/latest/testing.html#channeltestcase
The problem is, that the messages are not send when you call ```Channel(...).send(...)```. You have to call ```Channel(...).send(..., immediately=True)``` which is possible in one test, but not, if you test a function which calls the send message.

The problem is, that all messages are cached until ```pending_message_store.send_and_flush()``` is called. So I think it is a good idea to call it first thing in ```get_next_message```.

I added a test to test the example test. I did not know where the right place for this test is. I can move it to another file if you like.

### DIFF
--- a/channels/tests/base.py
+++ b/channels/tests/base.py
@@ -11,7 +11,7 @@ from django.test.testcases import TestCase, TransactionTestCase
 from .. import DEFAULT_CHANNEL_LAYER
 from ..asgi import ChannelLayerWrapper, channel_layers
 from ..channel import Group
-from ..message import Message
+from ..message import Message, pending_message_store
 from ..routing import Router, include
 from ..signals import consumer_finished, consumer_started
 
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I instead made the send delay functionality only work inside consumers in 69c59ee8b4924cd346e0a2b1724979375ad9924c, but thanks for the PR!

ostcar:
Thanks! That is much better. I hope you make a release with this fix soon, because with the current stable of channels our tests are broken and we either have to patch each test or stay on channels < 1.0

andrewgodwin:
Release should be coming tomorrow.

ostcar:
Wow. Great!

andrewgodwin:
Released in `1.0.2`

Issue #494, opened 2017-01-21
Add a ChannelLiveServerTestCase (similar to LiveServerTestCase)


bittner:
At the moment, for testing an application using channels we have two dedicated test case classes:

- [ChannelTestCase](https://github.com/django/channels/blob/master/channels/tests/base.py#L75-L76)
- [TransactionChannelTestCase](https://github.com/django/channels/blob/master/channels/tests/base.py#L79-L80)

For testing Django live applications (through the whole stack) Django provides two additional test case classes, which come  in handy especially in BDD (usually in combination with something like Selenium):

- [LiveServerTestCase](https://github.com/django/django/blob/master/django/test/testcases.py#L1262-L1272)
- [StaticLiveServerTestCase](https://github.com/django/django/blob/master/django/contrib/staticfiles/testing.py#L5-L11)

ChannelLiveServerTestCase
------------------

As for channels we have a completely different execution profile (we need interface servers and worker servers), yet still the same need for testing [features](http://behave.readthedocs.io/en/latest/gherkin.html), it would be helpful to have **a similar test case class for testing across the Channels stack**, e.g.

- `ChannelLiveServerTestCase`

Why In Channels?
--------------------

For obvious reasons (testing) it makes sense that this is integrated in Channels rather than a third-party package.

Examples Of Usage
--------------

We have discussed this in https://github.com/behave/behave-django/issues/16#issuecomment-274127381.

Setting up a test setup for use with Splinter or Selenium is pretty laborious, as [demonstrated in the same issue](https://github.com/behave/behave-django/issues/16#issuecomment-274117106) (comment above). A `ChannelLiveServerTestCase` would make that much easier, and the Django testing picture more complete.

andrewgodwin:
I agree it would be useful, but this falls outside my knowledge and experience range - I do not want to spend time getting tests working with WebSockets from browsers, they're finicky at the best of times. We can leave this open, though, as someone else may be interested.

bittner:
Thank you, @andrewgodwin, for agreeing!

It's not even about browsers at all. [LiveServerTestCase](https://github.com/django/django/blob/master/django/test/testcases.py#L1262-L1272) is [spinning up a runserver](https://github.com/django/django/blob/master/django/test/testcases.py#L1282-L1307) in the `setupClass` method of a `TransactionTestCase`. All server-side stuff.

[...]

PR #497, opened 2017-01-23
Add ChannelLiveServerTestCase

proofit404:
As we discussed earlier in #494 we need live server test case for example to run selenium against it.  This is work in progress PR, but I want some feedback already.

Few questions we need to answer before merging it.

1. Channels documentation mention test_channel_aliases as ability to test complex layers setup. I don't know what to do in this case for live server test. It is possible to spawn multiple live server instances. But its  not clear to me what user API we should provide.

2. Regular live server thread contains connection override logic. I don't understand right now what problem does it solve. It will be awesome if any one can describe the reason it present there or provide test case for it.

3. How we can validate that twisted bind port successfully? For now I select first port from the possible range without any validation.

4. apply_routes setup test instance with overriding setUp and tearDown methods. Live server spawns its threads in the test class setup (as original one do). So apply routes hook works after we start live server threads and overrides not available in those threads. For now I change apply_routes decorator to work on class basis. But we also can consider running live server for each individual test method however this will slow things down and inconsistent with django behavior.

5. We need to write docs.

cc @bittner


andrewgodwin:
1: I don't think there's a need for multiple aliases for this one - in fact, raise an error if someone tries to set it.

2: I don't know what it's for either.

3: I believe Twisted will traceback during listen if you try to listen on a port that's already bound. Best to verify this by trying to make it listen on something deliberately already used and checking what happens.

4: If we can make sure Daphne lives in a separate thread but the worker lives in the test thread somehow (though I'm not sure that's possible), the overrides would work. Otherwise, yes, we will need to at least restart the worker for each test.

I also think the test case class should be in a new module, not `base`.

bittner:
I can't comment on the implementation details related to Channels, because I haven't used it yet.

Some comments on the file organization:

- I agree, I'd move the test case classes into a module called `testcases` (just [like core Django](https://github.com/django/django/blob/master/django/test/testcases.py)).
- It's a bit irritating that we have the test infrastructure code _and_ the tests as such in one and the same directory. In core Django this seems more logical: the [tests are outside](https://github.com/django/django/tree/master/tests) of the `django` module, and the test infrastructure code lives in [django.test](https://github.com/django/django/tree/master/django/test) -- note the singular.

proofit404:
@bittner I absolutely agree with you. Its better to keep channel.test as primary package intended for users. We should provide `channels.tests` for backward compatibility which will show deprecation warning. Also channels own tests should live in the top tests directory and must be excluded from PyPI package. I'll do this change in separate PR if @andrewgodwin agree with this project layout.

@andrewgodwin This is my thoughts so far:

1. I agree we can make live server support for single channel layer only. If anyone needs multiple channel layers support for real, they will at least come up with some opinion about user api.

2. and 3. I'll research it at weekend.

4. Good point, I think it should solve this problem.

5. All django test cases lives in single module together with LiveServerThread class. This was my inspiration to put channels version into single base module. I don't mind to move it outside.

Thank for feed back guys!

bittner:
@andrewgodwin Do you agree with @proofit404's [thoughts on the `channels/tests/` directory](https://github.com/django/channels/pull/497#issuecomment-274842298), i.e. moving the actual tests out of the module, and create the test infrastructure in `channels/test/`, [as I suggested](https://github.com/django/channels/pull/497#issuecomment-274746740)?

Is there anything I can do to help this PR move forward?

proofit404:
@bittner sorry for the huge delay. I can continue working on this PR when I comeback from my business trip on 13th of February.

andrewgodwin:
I do think separating out tests and test infrastructure would be a good idea, yes.

proofit404:
Hello guys,

I stuck. Channels test cases use in memory layer implementation. Twisted reactor can be started and stopped only once during whole process live circle. If we use threads for live server, we can't write more than one test case at all. If we use multiprocessing for live server, we can't use in memory layer. `>_<`

Any suggestions?

proofit404:
Maybe running behave against real channels installation on developers machine is the best option here.

andrewgodwin:
Yes, I think given that limitation it might have to use a cross-process channel backend so we can run the server properly?

proofit404:
Yes, for live server test case we can substitute project settings with `asgi_ipc` instead of `asgiref.inmemory`. If you think this is proper solution of course.

andrewgodwin:
I don't want to encourage use of `asgi_ipc` to be honest, I'm considering deprecating it in favour of something else SQLite-based. I'd rather give people some way to feed in real channel layer settings.

proofit404:
It is possible to ignore channels test mixin. But that means we can't give certain guaranties about our environment. For example RabbitMQ layer doesn't provide flush extension. In this situation one test can influence into another through messages in the layer.

I can provide test mixin for RabbitMQ which must be used manually in each test case. But this wont be a general solution.

andrewgodwin:
Yeah, the flush is kind of important for tests. Maybe the test case could look for the flush extension and then error if it's not present? Both IPC and Redis provide it, so at least there's options.

proofit404:
Current RabbitmqChannelLayer create new virtual host for each test. This make guaranties that we do not overlap test data even during parallel execution. I can move this in to test helper and make it part of asgi_rabbitmq package. So if user doesn't inherit from this mixin in test, test will fail due to the lack of flush extension. If user inherit from this mixin, test will be executed in the sandbox environment.

Does this sound like a proper solution?

proofit404:
I changed live server implementation to multiprocessing.

Test suite currently half time randomly fails and half time randomly hangs. I don't know what is the reason for this random behavior. Probably it happens due flaky process communication.

isparks:
Not directly helpful to the code (sorry) but want to thank @proofit404 for continuing efforts on this.

proofit404:
Closed occasionally. Stupid buttons...

andrewgodwin:
I can't merge this in if the test suite is randomly failing, of course. Do you know why it might be, and if it's solvable? I'd propose there's no live tests in the standard Channels suite anyway, and that the live test case is just left for those with the will and the infrastructure set up to run it.

proofit404:
For now I only have some guesses.

Speaking about integration test suite. I don't mind if we move it out of the channels tests. But I have no idea what will be the right place for it.

andrewgodwin:
Well, I don't think that Channels needs integration tests built in right now; this ticket is just to provide the functionality for others to use, and that can live in `channels.test` without being used.

proofit404:
Ok, I'll try to test this manually. Maybe this weekend.

proofit404:
Hello guys,

The last `todo` in this PR relates to the port selection. If first port in the range is already bound, twisted reactor doesn't exit. As far as I understand `reactor.listenTCP` doesn't fail reactor in this case. This seems reasonable if we take into account that twisted reactor can be started and stopped only once.

Does `listenTCP` support errback passing? If not, we can check port availability manually before running reactor. Any suggestions?

proofit404:
Also I move `asgi_rabbitmq` test suite to this live server. This is how any one can use it with `asgi_rabbitmq` in their projects: https://github.com/proofit404/asgi_rabbitmq/blob/d0e5b388746999e2dccd51f87ac562f2aa190df0/tests/test_integration.py#L12

proofit404:
@andrewgodwin If you wish I can add same integration test to the `asgi_redis` so we will always sure that live server works for both official production layers.

andrewgodwin:
@proofit404 I don't think listenTCP does support errback, though I'm not that knowledgeable about Twisted internals.

As for the asgi_redis tests, I don't really mind either way; we wouldn't be able to run integration tests in Travis, of course, but it's probably nice to have them around in case people want to check?

proofit404:
Why we can't run them on Travis? asgi_rabbitmq run live server tests without any problem.

andrewgodwin:
I guess I had thought that wouldn't be possible, but if it is, the more the better.

channels/test/liveserver.py
```diff
@@ -113,6 +113,12 @@ def run(self):
 
 
 class ChannelLiveServerTestCase(TransactionTestCase):
+    """
+    Does basically the same as TransactionTestCase but also launches a
+    live Daphne server and Channels worker in a separate processes so
```
bittner: I think this should read "in a separate process, so that" (singular; comma).

channels/test/liveserver.py
```diff
@@ -113,6 +113,12 @@ def run(self):
...
+    that the tests may use another testing framework, such as Selenium
+    for example, instead of the built-in dummy client.
```
bittner: I think that should read "... test framework, such as Selenium, ..." (no -ing; no for example)

Just checking that we're PEP20 compliant! ("[Readability counts.](https://www.python.org/dev/peps/pep-0020/)")  :1st_place_medal: :smiley:

channels/test/liveserver.py
```diff
@@ -0,0 +1,205 @@
...
+from ..worker import Worker, WorkerGroup
+
+# NOTE: We use ChannelLayerManager to prevent layer instance sharing
+# between forked process.  Some layers implementations create
+# connections inside the __init__ method.  After forking child
+# processes can lose the ability to use this connection and typically
+# stuck on some network operation.  To prevent this we use new
+# ChannelLayerManager each time we want to initiate default layer.
+# This gives us guaranty that new layer instance will be created and
+# new connection will be established.
```
bittner: I know this may be seen as a stupid comment, but can you move lines 16-23 into a module docstring on top of the file? I think this gets read more often that way.

If it's really about the `from ..asgi import ChannelLayerManager` a single line comment on line 12 above should suffice (in addition, when the comment block is available as a docstring).

proofit404: I'm not sure about this change. This commend doesn't change module usage in any case. It is a reminder to the person who will change this module in the future.

channels/test/liveserver.py
```diff
@@ -0,0 +1,205 @@
...
+            else:
+                self.worker = WorkerGroup(
+                    channel_layer=channel_layers[DEFAULT_CHANNEL_LAYER],
+                    signal_handlers=False,
+                    n_threads=self.n_threads,
+                )
+            self.worker.ready()
+            self.is_ready.set()
+            self.worker.run()
+        except Exception:
```
proofit404: It was inspired with Django own LiveServerThread run method. I think it's reasonable.

proofit404:
The last know issue was resolved yesterday.

We don't need to check if port already bound. We simply bind zero port and relies on operation system to do proper thing. Exactly same change happens to Django master. I guess we can trust this technique.

This realization depends on https://github.com/django/daphne/pull/92 Please merge this PR after Daphne changes approve.

Can you guys give this live server additional testing? I already use it in the asgi_rabbitmq and asgi_redis integration test suite. But I don't want relies on my own testing only.

proofit404:
@andrewgodwin @bittner Any news? Did someone tries it?

bittner:
@proofit404 Let me ask. (I'm not a channels user, I have no project to try it, sorry.)

kradem:
Hi everyone!

I tried to test it in my project, but with limited success.

The problem is that it doesn't render css, but I realized that only after I stopped another `python manage.py runserver` process - it looks like that process had served css for this "separated" ChannelsLiveServerTestCase. And so I realized another potentially dangerous thing: it was using main database too while there was `python manage.py runserver` process alive, as test signup process was complaining about existing users (even database test_* was initially created and destroyed afterwards.

Btw, I don't use signup/login process in functional tests, but I couldn't make it with [this solution](http://stackoverflow.com/a/32708320) for ChannelsLiveServerTestCase.

proofit404:
Hi @kradem 

Thanks for your testing. Indeed worker subprocesses used to connect to real database instead of test mirror. I fixed that bug in the last commit. Can you check it again?

Unfortunately I can't reproduce static files serving bug (related to css in your case). Can you provide code sample?

Also I don't think `self.driver` from the stackoverflow answer uses server url provided by this test case. Can you point me to it implementation?

P.S. Current implementation fails with error on Django 1.8. I'll revisit it soon.

proofit404:
Django 1.8 compatibility was fixed.

kradem:
Hi @proofit404\!

The good news is that injected session works, but the bad news is it works only when there's no another running `python manage.py runserver` process.

I made a fork of channels-examples (without changing any existing code - DATABASES variable should be changed).

https://github.com/kradem/channels-examples/tree/liveserver-test-case/multichat/functional_tests

There's readme file explaining what you should do to prepare and run the test and also [https://github.com/kradem/channels-examples/blob/liveserver-test-case/multichat/functional_tests/screenshot%202017-03-24%2016-13-59.png screenshot] demonstrating the lack of css.

Please ask if you need anything else.

proofit404:
Hi @kradem 

Thanks for your example. It helps a lot! I fix issue with static files serving. Now live server test case serves static files correctly.

Also I successfully debug the root of troubles with session engine. In your example you use database backend for session storage. Also as I mention in the docs live server tests uses channel layer from settings. Together with workers from `django-admin runserver` command.

So we are in the situation where test worker can process messages from runserver Daphne. Or worker from runserver can process messages from test Daphne. Second variant is actually occurs in your test run. This happens because all messages comes through same redis db.

But workers created with runserver and tests have different settings. Default database is the main difference. So in your test you create cookie in the test database. Selenium tries to login. HTTP request was processed by worker process created with runserver. Consumer runs against multichat db, which has empty session store. We can't login because our session were created in the another database.

I change asgi layer to the rabbitmq and add `RabbitmqLayerTestCaseMixin` to the `MultichatFunctionalTest`. This mixin create new rabbitmq virtual host for each test. So this overlap doesn't happens. Test passed successfully with `django-admin runserver` running at the same time.

Channels does not has concept of test layers similar to test databases from Django. I mention this in docs. But now I clearly see it isn't enough. Many users will have similar problems and it is definitely hard to debug. Having `RabbitmqLayerTestCaseMixin` is helpful but to not a general solution.

I can develop same mixins for redis and ipc layers and force all users to setup at least one of it. But maybe it is time to introduce testing layer extension to asgi spec?

@andrewgodwin what do you think?

kradem:
Hi @proofit404\!

Yep, I successfully tested it with asgi_rabbitmq (and updated repo with it).

I'm looking forward seeing asgi_redis implementation too. If you think I may help please don't hesitate to ask for.

Regards!

proofit404:
Implementation of the same mixin for redis is trivial, but it requires `db` number config in the layer constructor. But before do that I want to know if we moving in the right direction.

Thanks for your testing!

proofit404:
@andrewgodwin Any news? Can I help somehow to speed up review process?

andrewgodwin:
I haven't had a chance to look at this yet, it's pretty big and the last couple of weeks have been very busy. I'll try to get to it soon.

channels/test/liveserver.py
```diff
@@ -0,0 +1,218 @@
...
+# between forked process.  Some layers implementations create
+# connections inside the __init__ method.  After forking child
+# processes can lose the ability to use this connection and typically
+# stuck on some network operation.  To prevent this we use new
+# ChannelLayerManager each time we want to initiate default layer.
+# This gives us guaranty that new layer instance will be created and
+# new connection will be established.
+
+
+class ProcessSetupMixin(object):
```
andrewgodwin: This doesn't need to be a mixin - both places it is used also inherit from `multiprocessing.Process`, so you can just make this inherit from that and have single inheritance.

docs/releases/1.1.0.rst
```diff
@@ -15,6 +15,9 @@ Major Changes
   work but will trigger a deprecation warning, and ``channels.tests`` will be
   removed completely in version 1.3.
 
+* Add ``ChannelLiveServerTestCase`` for integration testing.  It is
+  possible to run Selenium with JavaScript which operates WebSockets.
```
andrewgodwin: This now needs removing from these release notes as we're further along; don't worry about adding release notes, I'll do that when I make the release!

docs/testing.rst
```diff
@@ -294,3 +294,34 @@ mocked.
...
+
+In the test above Daphne and Channels worker processes were fired up.
+These processes run your project against the test database and the
+default channel layer you spacify in the settings.  If channel layer
+support ``flush`` extension, initial cleanup will be done.  So do not
+run this code against your production environment.  When channels
+infrastructure is ready default web browser will be also started.  You
+can open your website in the real browser which can execute JavaScript
+and operate on WebSockets.  ``live_server_ws_url`` property is also
+provided if you decide to run messaging directly from Python.
```
andrewgodwin: There's a few grammar and presentation fixes here but I'll fix those post-merge.

testproject/chtest/consumers.py
```diff
@@ -2,5 +2,6 @@
 
 
 def ws_message(message):
-    "Echoes messages back to the client"
+    """Echoes messages back to the client."""
```
andrewgodwin: Why was this changed? There's no need for `"""` unless you're doing multiple lines.

proofit404:
And we done. Any thoughts on testing extension or asgi_redis test mixin?

andrewgodwin:
What do you mean by those?

(also, it looks like master is still failing to test, so this probably will too - I'll fix that and then you can rebase)

proofit404:
Whole story starts from this [comment](https://github.com/django/channels/pull/497#issuecomment-288741729) @kradem provides additional testing and we made some conclusions. Testing extension were proposed to solve shared layer problem between test process and runserver process.

andrewgodwin:
Oh, you mean the idea of having a separate test channel layer to run tests on. I'd rather not make this an extension but have people configure a channel layer specifically for tests and pass its alias as an argument to the test class, I think. Trying to make it automatic is probably not possible on all backends, and if you want this level of advanced testing a bit more setup is fine.

proofit404:
Hooray test passes.

kradem:
Hi @proofit404\!

It works very well with the last asgi_rabbitmq, but it shows error in teardown process.

<details> 
  <summary>psycopg2.OperationalError: SSL error: sslv3 alert bad record mac</summary>
======================================================================
ERROR: test_multichat_functionality (functional_tests.test_channels_page.MultichatFunctionalTest)
----------------------------------------------------------------------
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/db/backends/utils.py", line 62, in execute
    return self.cursor.execute(sql)
psycopg2.OperationalError: SSL error: sslv3 alert bad record mac

The above exception was the direct cause of the following exception:

  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/test/testcases.py", line 216, in __call__
    self._post_teardown()
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/asgi_rabbitmq/test.py", line 63, in _post_teardown
    super(RabbitmqLayerTestCaseMixin, self)._post_teardown()
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/channels/test/liveserver.py", line 218, in _post_teardown
    super(ChannelLiveServerTestCase, self)._post_teardown()
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/test/testcases.py", line 908, in _post_teardown
    self._fixture_teardown()
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/test/testcases.py", line 943, in _fixture_teardown
    inhibit_post_migrate=inhibit_post_migrate)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/core/management/__init__.py", line 130, in call_command
    return command.execute(*args, **defaults)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/core/management/base.py", line 345, in execute
    output = self.handle(*args, **options)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/core/management/commands/flush.py", line 54, in handle
    allow_cascade=allow_cascade)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/core/management/sql.py", line 15, in sql_flush
    tables = connection.introspection.django_table_names(only_existing=True, include_views=False)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/db/backends/base/introspection.py", line 88, in django_table_names
    existing_tables = self.table_names(include_views=include_views)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/db/backends/base/introspection.py", line 56, in table_names
    return get_names(cursor)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/db/backends/base/introspection.py", line 52, in get_names
    return sorted(ti.name for ti in self.get_table_list(cursor)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/db/backends/postgresql/introspection.py", line 66, in get_table_list
    AND pg_catalog.pg_table_is_visible(c.oid)""")
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/db/utils.py", line 94, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/utils/six.py", line 685, in reraise
    raise value.with_traceback(tb)
  File "/home/balnir/.virtualenvs/multichat/lib/python3.5/site-packages/django/db/backends/utils.py", line 62, in execute
    return self.cursor.execute(sql)
django.db.utils.OperationalError: SSL error: sslv3 alert bad record mac

----------------------------------------------------------------------
</details>


The actual sql variable is "DROP DATABASE 'test_multichat'" command. That database is being dropped at the end of the test suite, but if I duplicate tests the other won't pass:

<details> 
  <summary>details</summary>
test_multichat_functionality (functional_tests.test_channels_page.MultichatFunctionalTest) ... ok
ERROR
test_multichat_functionality2 (functional_tests.test_channels_page.MultichatFunctionalTest) ... ERROR
ERROR
</details>

Regards!

andrewgodwin:
I just tried this against a 1.11 installation and it looks like `.is_in_memory_db()` has gone away - if you merge against master again you will pull in https://github.com/django/channels/commit/9266521f9445e4431ae3567a5e2ce25b6af9133c, which does run them against 1.11, but it seems the error is not actually triggered until you try to run tests with the actual test case. Running a test works fine on 1.10, though.

I remember we discussed running integration tests in Travis a while back - did I say no, or was there a technical reason we rejected it?

proofit404:
Turns out `is_in_memory_db` call signature was changed in Django 1.11. I already fix this.

Regarding integration tests we decided to exclude them into channel layers. Running this tests requires real layer implementation. Because of that it go against channel policy not use any layer package as a dependency for channels.

I use this live server test cases for integration testing in RabbitMQ layer.

Also there is an open PR in Redis layer.  But you can't merge it yet because of [this lines](https://github.com/django/asgi_redis/pull/32/files#diff-354f30a63fb0907d4ad57269548329e3R14) before next channels release.

Ideally we need to run layers tests against channels master after each push to it. Probably we can do it with TravisCI API. Running special script in the `after_build` section should be sophisticated solution.

jricardo27:
Hi, thanks for the great work.

I wonder if are there any plans for defining a `ChannelStaticLiveServerTestCase` ala `django.contrib.staticfiles.testing.StaticLiveServerTestCase` ?

andrewgodwin:
We can definitely get Travis running the layer tests as well after every commit, using the Redis library. I'm going to merge this, and then the things left I'd like one of us to look at are:

- Having a test alias for the channel layer, and possibly forcing you to define one if you want to use the liveserver test
- Staticfiles serving support (probably as a class option rather than another subclass)
- Getting Travis testing integration tests with this as well

bittner:
Wahoo! :astonished: This is merged, great! :tada:

Thanks @proofit404, thanks thanks thanks!

(Now let's make sure this stay stable and usable for everyone.)

proofit404:
Hi @andrewgodwin 

I'm glad we merge it! Regarding further work on this test case:
* I think test alias is good idea. This way we need special support only in the layers which doesn't support `flush` extension.
* Static files serving support [already here](https://github.com/django/channels/blob/master/channels/test/liveserver.py#L86). We can add `serve_static` boolean flag similar to `worker_threads` class attribute. I think this should be enabled by default. Most people expect this to work out of the box.
* I prefer to keep integration tests inside layers repositories. I already extend this test case in the rabbitmq layer for advanced logging and profiler support inside child process. Channels definitely doesn't need to know about this internal machinery. Also each layer can have different set of integration tests. Lets see what TravisCI is able to do for non monorepo projects.

First two items of this list already clean enough. I'll implement them.

andrewgodwin:
Ah, yes, I missed the static stuff. We should have a way to turn it on/off easily for sure as people might not want static files on.

proofit404:
So far we have
* #604 implements test aliases
* #603 implements serve_static flag
* I found https://github.com/ajdm/travis-build-children instructions. But it requires ownership privilege so I pass this to you @andrewgodwin :)

UrAvgProgrammer:
Hi @proofit404, I am currently using ChannelsLiveServerTestCase to spin up a runserver for behave, I am using StaticLiveServerTestCase at first but have to change to ChannelsLiveServerTestCase for the websocket support but for some reason now my test for my upload-feature is failing. Does ChannelsLiveServerTestCase support uploading file like StaticLiveServerTestCase?

proofit404:
Hi @UrAvgProgrammer 

I left channels project a long time ago. I can't even recall what all those things have to mean. Sorry :disappointed: 

Regards, Artem.

Issue #498, opened 2017-01-23
Channels fails on Windows server


LegoStormtroopr:
The error is being raised from twisted, but all three test runs use `Twisted==16.6.0`.
I'm not 100% sure this is a channels specific thing, but thats what raised the issue.

I'm going to do some more investigation, but would the project be interested in including an `appveyor.yml` to add Windows testing file if one was put together?

If you're submitting a bug report, please include:

- Your OS and runtime environment, and browser if applicable - **Windows server 2016, Python 3**
- The versions of Channels, Daphne, Django, Twisted - **see below**
- What you expected to happen vs. what actually happened - **channels runserver fails to start**
- How you're running Channels (runserver? daphne/runworker? Nginx/Apache in front?) - **runserver**
- Console logs and full tracebacks of any errors - **see below**

```
Traceback (most recent call last):
  File "manage.py", line 15, in <module>
    execute_from_command_line(sys.argv)
  File "C:\Python35\lib\site-packages\django\core\management\__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "C:\Python35\lib\site-packages\django\core\management\__init__.py", line 328, in execute
    django.setup()
  File "C:\Python35\lib\site-packages\django\__init__.py", line 18, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "C:\Python35\lib\site-packages\django\apps\registry.py", line 115, in populate
    app_config.ready()
  File "C:\Python35\lib\site-packages\channels\apps.py", line 22, in ready
    monkeypatch_django()
  File "C:\Python35\lib\site-packages\channels\hacks.py", line 10, in monkeypatch_django
    from .management.commands.runserver import Command as RunserverCommand
  File "C:\Python35\lib\site-packages\channels\management\commands\runserver.py", line 5, in <module>
    from daphne.server import Server, build_endpoint_description_strings
  File "C:\Python35\lib\site-packages\daphne\server.py", line 6, in <module>
    from twisted.internet.endpoints import serverFromString
  File "C:\Python35\lib\site-packages\twisted\internet\endpoints.py", line 36, in <module>
    from twisted.internet.stdio import StandardIO, PipeAddress
  File "C:\Python35\lib\site-packages\twisted\internet\stdio.py", line 30, in <module>
    from twisted.internet import _win32stdio
  File "C:\Python35\lib\site-packages\twisted\internet\_win32stdio.py", line 9, in <module>
    import win32api
ImportError: No module named 'win32api'
```

[...]

PR #502, opened 2017-01-24
Add appveyor support to add windows tests

LegoStormtroopr:
As per issue #498 here is a PR to add appveyor support to test windows compatibility.

There is only one change in the channels code - switching multiprocessing for psutil when calculating the cpu_count , as it caused a NotImplementedError on Python 2.7 on windows. I'm happy to revert that if you only want windows testing on Python 3.
https://docs.python.org/2/library/multiprocessing.html#multiprocessing.cpu_count

I also added a small util library (`toxmatch.py`) that does the same as travis-tox for appveyor to get the list of allowable test environments that can be run from a tox.ini matrix 

I've also made some small tweaks to `tox.ini` to reduce the number of lines and allow explicit platform specific testing.

For this to work someone on the team will need to sign up at Appveyor to configure hooks and such. But you can confirm it works smoothly here - https://ci.appveyor.com/project/LegoStormtroopr/channels

Enjoy your verified Windows compatibility and [brand new badge!](https://www.appveyor.com/docs/status-badges/)

For details on starting redis see: http://stackoverflow.com/questions/28673763/start-redis-server-on-appveyor

Some of this was influenced by the appveyor.yml from the Twisted repository
https://github.com/twisted/twisted/blob/trunk/appveyor.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,5 @@ python:
   - "3.4"
   - "3.5"
 install: pip install tox-travis
-script: tox
+script:
+  - tox
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,6 +22,4 @@ install:
 
 # command to run tests
 test_script:
-  - cmd: python ci/toxmatch.py > tox.env.txt && set /p TOXENV=<tox.env.txt
-  - cmd: python ci/toxmatch.py
-  - cmd: tox -e %TOXENV%
+  - cmd: tox --skip-missing-interpreters
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,27 @@
+#
+# Some help for Python on AppVeyor:
+# * https://packaging.python.org/appveyor/
+# * https://ci.appveyor.com/tools/validate-yaml
+
+build: false
+
+environment:
+  matrix:
+      - PYTHON_HOME: "C:/Python35"
+      - PYTHON_HOME: "C:/Python34"
+      - PYTHON_HOME: "C:/Python27"
+
+install:
+  - choco install redis-64
+  - redis-server --service-install
+  - redis-server --service-start
+  - cmd: |
+        set PATH=%PYTHON_HOME%;%PYTHON_HOME%\Scripts;%PATH%
+        python -m pip install --upgrade pip
+        pip install tox
+
+# command to run tests
+test_script:
+  - cmd: python ci/toxmatch.py > tox.env.txt && set /p TOXENV=<tox.env.txt
+  - cmd: python ci/toxmatch.py
+  - cmd: tox -e %TOXENV%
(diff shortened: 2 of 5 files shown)


andrewgodwin:
Is there a reason toxmatch needs to be in the top level directory? I'd rather have it somewhere else (I'm also surprised something like this doesn't already exist!)

You'll also have to fix up Jenkins, it looks like it's trying to run the Windows variant of all the tests (when it seemingly can't)

LegoStormtroopr:
No reason and it can live anywhere. I wasn't sure where, so I figured I wait for you to ask that question and move it accordingly. Tox is implementing a wildcard approach, but I'm happy to pull this out into a separate tool (because I need it too), but that will take longer.

Its odd that the windows tests are being run - they shouldn't be and tox should be stopping that as platform dependent switching is built in. That may be because `travis-tox` does some odd monkey-patching to implement the `toxmatch`-like functionality ([their words, not mine](https://github.com/ryanhiebert/tox-travis/blob/master/src/tox_travis/after.py#L27))

As for Jenkins, I'll have a go. Did you mean travis, or is there more testing that I missed?

andrewgodwin:
Urgh, I meant travis, not jenkins, my fault. Work leaking through.

I don't mind if it lives in the repo, but I do think it would be nicer if it worked as travis-tox does, rather than adding slightly irrelevant code to the repo (though if it was in a `ci/` subdirectory I wouldn't mind much)

LegoStormtroopr:
Ah cool. Wasn't sure. I agree about having it work like travis-tox does is better long term, but `ci/` seems like an good immediate improvement.

I should have a fix for the travis stuff tomorrow.

LegoStormtroopr:
Hooray! It finally works, everywhere. Congrats on your Windows support. https://ci.appveyor.com/project/LegoStormtroopr/channels/build/1.0.53

Krukov:
`psutils` just for determine minor value in one place for python2.7 at windows? Maybe we can catch `NotImplementedError` (or something like this without new dependency)? 
Also `psutil.cpu_count()` can return `None`

LegoStormtroopr:
The easy alternative is to revert to multiprocessing and drop python 2 support on windows. If the code catches NotImplemented it will still need to determine (or fake) the cpu_count.

proofit404:
We can use multiple environment markers in the requirements.txt file used in tox. After that we can wrap import in the if statement with similar environment markers.

LegoStormtroopr:
Worker.py has been cleaned up and psutils has been removed as a requirement.

https://ci.appveyor.com/project/LegoStormtroopr/channels/build/1.0.56

LegoStormtroopr:
Is there a plan to pull this in? Happy to update the branch, but it went real quiet

andrewgodwin:
Oh, I just forgot about it. Updating it would be nice? I just need to find the time to set up the rest of the Appveyor stuff I guess, I'm not super concerned about Windows testing at the moment.

andrewgodwin:
Sorry, left this open for so long - going to just close it for now, hopefully I'll get a chance to look at appveyor support another time, but we've not had any Windows-specific issues so far so it's not high on my todo list.

PR #507, opened 2017-01-25
Inbound Binding validation

Krukov:
Regarding #377 and https://github.com/django/channels/pull/469

Added `validate` method for websocket binding.

### DIFF
--- a/channels/binding/base.py
+++ b/channels/binding/base.py
@@ -11,6 +11,8 @@ CREATE = 'create'
 UPDATE = 'update'
 DELETE = 'delete'
 
+ALL_FIELDS = '__all__'
+
 
 class BindingMetaclass(type):
     """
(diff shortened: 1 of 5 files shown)


andrewgodwin:
I don't see what the error reporting mechanism here is - could you explain it a bit?

Krukov:
Uhhh you right. ) I think we don't need  reporting mechanism itself. I'll try to explain my vision. 
As I realize, Websocket protocol have not any specific way to notify clients about errors or something similar ( although we can initialize `close` message with status code and reason, but it is not suitable solution) (Am I right?). 
So we can send a message with some flag (like `{'status': 'error', ...`), but a client must understand this flag... We already have mechanism for sending messages (thank you captain obvious), but I don't think that we must dictate our standard for errors messages or implement a generic way to do this. 
`validate` method it is just a single entrypoint to determine behavior at valid/invalid data.  If you want to send errors to a client you can use your own `ChannelSocketException` and raise it at this method. 
Example for form validation mixin:

```python
class InvalidDataError(ChannelSocketException):

    def __init__(self, errors):
        self._message = json.dumps({'errors': errors})

    def run(self, message):
        message.reply_channel.send({'text': self._message})


class FormValidationMixin(object):
    form = None

    def validate(self, data, instance=None):
        data = super().validate(data, instance)
        form = self.get_form_class()(data=data, instance=instance)
        if not form.is_valid():
            raise InvalidDataError(form.errors)
        return form.cleaned_data

    def get_form_class(self):
        return self.form or modelform_factory(self.model, fields=self.fields, exclude=self.exclude)
```

andrewgodwin:
Well, yes, the main thing I really want to address is having a standard error reporting mechanism - that's the hard part here. As I said before, I'd rather have form validation in a third party package and just put the hooks into Channels to allow validation to run and errors to be sent.

Krukov:
@andrewgodwin  Sorry for a long response. I close this PR. I'll try to remake it as soon as I can.

PR #515, opened 2017-01-30
Simplify testing infrastructure

maikhoepfel:
This is the last of a round of pull requests to all Channels projects, to ensure all projects run tests with tox and Travis. Channels was mostly there, so I only had to bring it more in-line with the other projects.

This pull requests removes a layer of indirection, `tox-travis`. While it's a good and useful project, I think it makes it harder to reason about what Travis will execute, and seems to have some problems. Please see my commit message for an explanation. Furthermore, there's value in doing the same thing we do in the other four Channels projects.


andrewgodwin:
Much cleaner, thanks.

PR #516, opened 2017-01-31
Fix testing infrastructure.

KhasanovBI:
Changes:
1. Move requirements for tests in separate file. I think making extras tests is a bad idea, because extra have another sense to install additional functionality to current package, that not always need. For example in requests we can install `pip install requests[security]` or `pip install requests[socks]`. Tests requirements need for channels developers, not for programmers that using channels.
2. Fix tox envlist. Here is mistake in tox envlist.
3. Add python3.6 to travis. What about new Python 3.6? :)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 envlist =
-    {py27,34,35}-django-{18,19,110}
-    {py27,py35}-flake8
+    py{27,34,35}-django-{18,19,110}
+    py{27,35}-flake8
     isort
 
 [testenv]


KhasanovBI:
#515

andrewgodwin:
Why move tests into a requirements file? The extras solution works fine and eliminates having to rely on yet another file.

The python 3.6 stuff I agree with.

KhasanovBI:
IMHO it's like incapsulation of inner functionality from developers that use channels. I'm revert this change, if you didn't agree.

andrewgodwin:
Yes, I'd rather keep it like it is, please.

KhasanovBI:
@andrewgodwin done.

maikhoepfel:
Hey @KhasanovBI, thanks for the PR. 

I agree that declaring test dependencies in `setup.py` doesn't feel ideal, but then I don't think any of the available options does - and in the end, it does not matter much, and it's more important that they're declared at all. FWIW, `sentry-raven` and `pip` itself do it this way, which swayed me to that solution.

Good catch on the envlist! Not sure how I didn't catch that, thanks.

Regarding Python 3.6: Personally I would wait until we support a Django version which supports Python 3.6, and generally would update dependencies in lock step with Django releases. I think having a clear message (such as "we support all popular Python versions that Django currently supports") is more important than supporting the (rare?) use case of users wanting to run channels with Django master. 

But I'm 0-, so it's no blocker. But if we declare 3.6 support here, we should update all other Channels projects as well.

andrewgodwin:
I agree on not having 3.6 for now, it's nice to match Django. It's not hard to add when we need to (and I'm not aware of any big differences that would make code work on 3.5 but not 3.6)

KhasanovBI:
@andrewgodwin done.

PR #528, opened 2017-02-14
Update Binding to support models with UUIDField as primary key

pchiquet:
'update' message was sent when creating a new model instance because `instance.pk` is set for non auto pk.

The `instance.pk is None` check to detect creating instances only works for auto PKs.


andrewgodwin:
Nice, thanks!

PR #530, opened 2017-02-14
Scheduler Protocol Server

devmapal:
This is a first implementation for #74 

The implementation follows a similar design as the `channels.delay` app by providing a worker that listens on 'asgi.schedule'. Internally, apscheduler is used for scheduling the jobs, providing an apscheduler interface via django channels. The API design is inspired by https://github.com/johndoee/txasgiresource#scheduler.

If it is not desired to pull apscheduler in as a dependency, I could implement a similar functionality directly in the `channels.scheduler` app.

I have not yet written any documentation yet as I would hold off with that until message format is accepted/agreed on.


devmapal:
The test on travis-ci are currently failing for python 3.5. Unfortunately I am not able to reproduce this problem locally, for me all tests pass.

devmapal:
I was able to reproduce the issue locally now and fix it with the latest commit 40f1a42

andrewgodwin:
Interesting; I never saw a scheduler server as listening on a channel for jobs, but rather that it would have a config file (like cron), or just be controlled via the database and a GUI.

How do you envision people setting up jobs and being able to deploy them to different environments?

devmapal:
Sorry for the late reply.

I agree that providing a channels interface adds more complexity without providing much benefit. As the jobs are anyway already stored in the database, I think it would make more sense to provide manager methods for adding and removing jobs as well as an admin interface similar to [django_celery_beat](https://github.com/celery/django-celery-beat) for managing jobs.

Providing a configuration file could be an alternative approach for deploying scheduler configuration.

andrewgodwin:
Yes, that seems like a sensible set of interfaces; it would be OK to merge with just a config file and manager methods, I think, an admin UI is nice but not super required.

devmapal:
Celery beat makes the assumption that there is always only one scheduler, as otherwise locking is required to prevent jobs from being executed multiple times. Implementing those locks would be tricky especially for a file-based configuration.
I would however argue that it is desirable to allow multiple schedulers to allow more reliable job scheduling.
Do you agree with this (allowing multiple schedulers)? The consequences are that apscheduler can't be used and a custom scheduler + locking needs to be implemented.

andrewgodwin:
Allowing multiple schedulers is incredibly difficult; you can try it if you want, but it's a distributed systems problem that might take a long time to solve.

devmapal:
Closing this for now as I unfortunately don't have the time to continue working on it at the moment

domino14:
cmon bro

Brandl:
What is the current status on this? 
I was looking for such a feature and am willing to dedicate time to make it happen.

andrewgodwin:
It's currently not being worked on, though I might take a look at it as part of the big refactor for channels 2.

Brandl:
Is there anything I can help you with?

PR #531, opened 2017-02-15
Separate tests into own directory.

proofit404:
As we discussed in many issues it is necessary to separate own channels tests from channels test framework.

Also I rename channels.tests package into channels.test similar to Django. This change **is** backward compatible.


andrewgodwin:
This seems good generally, just need to get Travis happy I think.

proofit404:
It was old pip on Travis.

andrewgodwin:
Good, thanks for this!

Issue #451, opened 2016-12-13
My routing and consumer stops responding to incoming messages during high traffic


Key1337:
Is there some reason why my routing and consumer would stop working (listening) to incoming websocket messages after a period of high traffic (lots of incoming messages)?

I have an application that works correctly and I use it for sending and receiving websocket messages. Whenever I greatly increase the rate of incoming messages, eventually the routing & consumer stop responding. These messages are coming into the server from a non-local host. I can tell the messages are still being received at the front end by monitoring redis with this terminal command:

     redis-cli monitor|grep -v EVALSHA | grep -v LPOP

When this condition occurs, I can still send outgoing websocket messages from my django channels application. That code is in another area not included here. This is all occurring over the same websocket connection which never drops.

routing.py

    from django.db import transaction
    from channels.routing import route, route_class
    from channels.generic.websockets import WebsocketDemultiplexer
    from channels.sessions import channel_session, enforce_ordering
    from one_database.models import Server
    import logging

    logger = logging.getLogger(__name__)

    from one.consumers import command_gRQAXFR

    class GameDemultiplexer(WebsocketDemultiplexer):
        mapping = {
            "gRQAXFR": "gRQAXFR.receive",
        }
        http_user = True
        strict_ordering = True

        @transaction.atomic
        @channel_session
        def connect(self, message, **kwargs):
            logger.info("\n[one] GameDemultiplexer: Connected")
            skey = kwargs["server_key"]
            message.channel_session["server_key"] = skey
            logger.info("Server key:  " + skey)
            sip = dict(message.content["headers"])[b'x-forwarded-for'].decode("utf8")
            logger.info("Server IP:   " + sip)
            try:
                srv = Server.objects.select_for_update().get(server_key=skey, server_ipaddress=sip)
                logger.info("Server Name: " + srv.server_name)
                srv.server_game_channel=message.reply_channel
                srv.save()
            except Server.DoesNotExist:
                logger.info("\n[one] GameDemultiplexer: Server not found!")
                self.close()
            pass
        
        @transaction.atomic
        def disconnect(self, message, **kwargs):
            logger.info("\n[one] GameDemultiplexer: Disconnected")
            skey = kwargs["server_key"]
            logger.info("Server key:  " + skey)
            try:
                srv = Server.objects.select_for_update().get(server_key=skey)
                logger.info(srv.server_name)
                srv.server_game_channel=""
                srv.save()
            except Server.DoesNotExist:
[...]

PR #532, opened 2017-02-17
Requeue next message immediately to avoid wait queue race condition

Coread:
There is a race condition in the enforce_ordering mode. If the currently running message is processed quickly then it would attempt to re-queue messages from the wait queue before the next out of order message has been added to it. You can reproduce this condition simply by putting a sleep before adding the message to the wait channel.

Could potentially be the cause of #451

Fixed by re-queuing the next message immediately to the original channel, rather than the wait channel.

### DIFF
--- a/channels/sessions.py
+++ b/channels/sessions.py
@@ -123,10 +123,14 @@ def enforce_ordering(func=None, slight=False):
                         break
             else:
                 # Since out of order, enqueue message temporarily to wait channel for this socket connection
-                wait_channel = "__wait__.%s" % message.reply_channel.name
-                message.content["original_channel"] = message.channel.name
+                # If this message is going to be next, requeue it immediately
+                if order == next_order + 1:
+                    channel = message.channel.name
+                else:
+                    channel = "__wait__.%s" % message.reply_channel.name
+                    message.content["original_channel"] = message.channel.name
                 try:
-                    message.channel_layer.send(wait_channel, message.content)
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Hmm, the reason I moved from the method this diff proposed to a wait channel in the first place now eludes me, but I know there was reasoning behind it. Let me take a bit and work through this to make sure it's alright.

Coread:
Was the original change to use a wait queue to avoid workers spinning? If the current request took a long time to execute then all pending requests would continually resubmit themselves to the original channel.

andrewgodwin:
Yes, I think that might have been it, the idea being they're shuffled off into the wait channel and not re-examined until the next message in order comes along, rather than spinning around on the queue.

I wonder if there is a way to solve both problems at the same time, potentially by double-consuming the wait channel somehow. In particular, if I understand correctly, the problematic sequence here is:

* Message 3 comes in and is considered out of order by some code
* Message 2 comes in and processes, wait queue is flushed
* Message 3 is added to the wait queue

It feels to me like re-checking the session to see if a message has become current after it's added to the wait list might do it?

Coread:
Latest attempt re-queues the messages back onto the original channel in the event that the next_order has changed to the current message order while it was being processed.

andrewgodwin:
Yes, I think I like this approach. Ready to merge it in?

Coread:
yeah go ahead

andrewgodwin:
Thanks!

Issue #534, opened 2017-02-17
Allow a custom json [de]serializer in JsonWebsocketConsumer


dougkeen:
Currently, `JsonWebsocketConsumer`, `WebsocketMultiplexer`, and `WebsocketDemultiplexer` have `json.dumps` and `json.loads` coded into their behavior in a non-extensible way. It would be useful to be able to configure something like this:
```Python
class MyCustomJsonConsumer(JsonWebsocketConsumer):
 
    # Incoming messages get snake case keys (Python standard)
    json_loader = lambda text: underscorize(json.loads(text))

    # Outgoing messages get camel case keys (Javascript standard)
    json_printer = lambda content: json.dumps(camelize(content))

    ...
```

dougkeen:
It's also worth noting that a memoized custom serializer would alleviate the inefficiencies identified in #474

PR #535, opened 2017-02-17
Allow custom json encoder and decoder in `JsonWebsocketConsumer` (#534)

dougkeen:
Implements feature request #534

### DIFF
--- a/channels/generic/websockets.py
+++ b/channels/generic/websockets.py
@@ -139,6 +139,14 @@ class WebsocketConsumer(BaseConsumer):
         pass
 
 
+def _default_json_decoder(text):
+    return json.loads(text)
+
+
+def _default_json_encoder(content):
+    return json.dumps(content, cls=DjangoJSONEncoder)
+
+
 class JsonWebsocketConsumer(WebsocketConsumer):
     """
     Variant of WebsocketConsumer that automatically JSON-encodes and decodes
@@ -146,9 +154,15 @@ class JsonWebsocketConsumer(WebsocketConsumer):
     error on binary data.
     """
 
+    # Optionally configure a custom json decoder
+    json_decoder = staticmethod(_default_json_decoder)
+
+    # Optionally configure a custom json encoder
+    json_encoder = staticmethod(_default_json_encoder)
+
     def raw_receive(self, message, **kwargs):
         if "text" in message:
-            self.receive(json.loads(message['text']), **kwargs)
+            self.receive(self.json_decoder(message['text']), **kwargs)
         else:
             raise ValueError("No text section for incoming WebSocket frame!")
 
@@ -162,11 +176,11 @@ class JsonWebsocketConsumer(WebsocketConsumer):
         """
         Encode the given content as JSON and send it to the client.
         """
-        super(JsonWebsocketConsumer, self).send(text=json.dumps(content), close=close)
+        super(JsonWebsocketConsumer, self).send(text=self.json_encoder(content), close=close)
 
     @classmethod
-    def group_send(cls, name, content, close=False):
-        WebsocketConsumer.group_send(name, json.dumps(content), close=close)
+    def group_send(cls, name, content, close=False, json_encoder=_default_json_encoder):
+        WebsocketConsumer.group_send(name, json_encoder(content), close=close)
 
 
 class WebsocketDemultiplexer(JsonWebsocketConsumer):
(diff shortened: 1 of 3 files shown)


channels/generic/websockets.py
```diff
@@ -162,11 +176,11 @@ def send(self, content, close=False):
         """
         Encode the given content as JSON and send it to the client.
         """
-        super(JsonWebsocketConsumer, self).send(text=json.dumps(content), close=close)
+        super(JsonWebsocketConsumer, self).send(text=self.json_encoder(content), close=close)
 
     @classmethod
-    def group_send(cls, name, content, close=False):
-        WebsocketConsumer.group_send(name, json.dumps(content), close=close)
+    def group_send(cls, name, content, close=False, json_encoder=_default_json_encoder):
```
andrewgodwin: This can be gotten from `cls`, surely, rather than being passed in?

channels/generic/websockets.py
```diff
@@ -249,27 +266,28 @@ class WebsocketMultiplexer(object):
     stream = None
     reply_channel = None
 
-    def __init__(self, stream, reply_channel):
+    def __init__(self, stream, reply_channel, json_encoder=_default_json_encoder):
```
andrewgodwin: I'd rather this was a None value, and it directly fell back to json.dumps rather than a global variable

andrewgodwin:
Mostly fine, but can the default implementations please just be on the class directly and not new global variables in the module? Most of the cases where they're used directly can either be accessed using `cls` or just use `json.dumps` directly

channels/generic/websockets.py
dougkeen: Oops, totally overlooked that. Good call.

dougkeen:
👍 Will do. High five for quick feedback!

channels/generic/websockets.py
```diff
@@ -243,33 +268,36 @@ class WebsocketMultiplexer(object):
...
     The multiplexer object is passed as a kwargs to the consumer when the message is dispatched.
-    This pattern allows the consumer class to be independant of the stream name.
+    This pattern allows the consumer class to be independent of the stream name.
     """
 
     stream = None
     reply_channel = None
 
-    def __init__(self, stream, reply_channel):
+    def __init__(self, stream, reply_channel, json_encoder=None):
```
dougkeen: Had to keep this as a constructor arg/instance var, since there's no way to propagate the demuxer's class variable into the `WebsocketMultiplexer` _class_, and the demuxer directly instantiates this class so there's no existing way to inject a custom `WebsocketMultiplexer` class.

I suppose we could make a class var in `WebsocketDemultiplexer` to hold a custom multiplexer class, but then the developer would have to remember to define custom json encoders/decoders on custom `WebsocketDemultiplexer` _and_ `WebsocketMultiplexer` subclasses, _and_ remember to hook up the custom `WebsocketMultiplexer` subclass to the custom `WebsocketDemultiplexer` class. The code here plays a little hot potato with the `json_encoder`, but it feels like it would provide an easier API in practice.

channels/generic/websockets.py
```diff
@@ -146,9 +146,19 @@ class JsonWebsocketConsumer(WebsocketConsumer):
     error on binary data.
     """
 
+    # Optionally configure a custom json decoder
+    json_decoder = None
+
+    # Optionally configure a custom json encoder
+    json_encoder = None
```
andrewgodwin: Why not have these as functions that just implement `json.loads` and `json.dumps`, then you can drop the if statements below?

channels/generic/websockets.py
andrewgodwin: Yeah, I see the problem here about propagating it; this is the one thing on my first scan through that I sort of didn't like but understood why it had to be like this. Arguably it's the fault of the existing code architecture for not being so great.

The problem is that this way does end up making `group_send` take the extra argument and you'd have to pass it directly through from the calling code, which is not ideal. I might prefer the whole "specifying the multiplexer subclass on the demultiplexer" thing in that case.

channels/generic/websockets.py
```diff
@@ -162,11 +162,58 @@ def send(self, content, close=False):
...
+    def encode_json(cls, content):
+        return json.dumps(content)
 
     @classmethod
     def group_send(cls, name, content, close=False):
-        WebsocketConsumer.group_send(name, json.dumps(content), close=close)
+        WebsocketConsumer.group_send(name, cls.encode_json(content), close=close)
+
+
+class WebsocketMultiplexer(object):
```
dougkeen: FYI, this is almost entirely just a move of this class (since it's referenced inside `WebsocketDemultiplexer`)...

channels/generic/websockets.py
```diff
@@ -162,11 +162,58 @@ def send(self, content, close=False):
...
+    def encode(cls, stream, payload):
+        """
+        Encodes stream + payload for outbound sending.
+        """
+        content = {"stream": stream, "payload": payload}
+        return {"text": cls.encode_json(content)}
+
+    @classmethod
+    def group_send(cls, name, stream, payload, close=False):
+        message = cls.encode(stream, payload)
```
dougkeen: ... except I changed `WebsocketMultiplexer.encode` to `cls.encode` here to keep things polymorphic.

andrewgodwin:
Thanks - sorry I forgot to merge it for a bit there!

PR #537, opened 2017-02-20
use domain instead of port

HassenPy:
i think this is a typo, as cookies aren't restricted by port, so even if you offload on the same domain, daphne will still work without having to specify a ``session_key`` GET parameter since it can read the django session cookie which it defaults to.


andrewgodwin:
Yes, I think I was probably relaying old info there. Thanks.

PR #539, opened 2017-02-20
use domain instead of port final

HassenPy:
just noticed there's other occurrences, this should be it.
the commits look ugly thou, just done with github interface, lazy to jump to terminal ..., can redo later if u want.


andrewgodwin:
It's fine, I squashed it down into a single commit to merge. Thanks.

PR #541, opened 2017-02-23
Set self.kwargs in Binding.trigger_inbound when setting self.message

pchiquet:
- it is useful to generate dynamic group names based on the path
- it is consistent with BaseConsumer that saves both message and kwargs


andrewgodwin:
Thanks!

PR #544, opened 2017-02-24
add Channels WebSocket javascript wrapper

fcurella:
This implements the Channels WebSocket javascript wrapper, mentioned in https://channels.readthedocs.io/en/stable/binding.html#frontend-considerations

The wrapper is available via staticfiles and it's been published to [npm](https://www.npmjs.com/package/django-channels). Users can use the included static file to send and receive messages, or extend the class provided by the npm module to implement the behaviour they need for their favorite framework.

### Examples of usage

First, you must include the javascript library in your template:

```
{% load staticfiles %}

{% static "channels/js/websocketbridge.js" %}
```

To process messages:

```
import { WebSocketBridge } from 'django-channels'

const webSocketBridge = new channels.WebSocketBridge();
webSocketBridge.connect();
webSocketBridge.listen(function(action, stream) {
  console.log(action, stream);
});
```

To send messages:

```
webSocketBridge.send({prop1: 'value1', prop2: 'value1'});

```

To demultiplex specific streams:

```
const webSocketBridge = new channels.WebSocketBridge();
webSocketBridge.connect();
webSocketBridge.listen();
webSocketBridge.demultiplex('mystream', function(action, stream) {
  console.log(action, stream);
});
webSocketBridge.demultiplex('myotherstream', function(action, stream) {
  console.info(action, stream);
});
```

To send a message to a specific stream:

```
webSocketBridge.stream('mystream').send({prop1: 'value1', prop2: 'value1'})
```

### Things I'm still not sure about

* I'd like for the npm module and channels to share the same version number. I'm not sure about you release process, and what could make it easier for you to keep the version number in-sync in two different files easily.
* Documentation for the JS API can be generated with `npm run docs`. I haven't included it for readability sake. I'm also not sure if or how we should host the JS API docs anywhere.


andrewgodwin:
So, this looks very fully featured and I am particularly fond of the fact it comes with docs.

However (and this may be a stupid question, please humour me), it looks like the code is duplicated into two places; one copy in `static` and one in `node`. Is it duplicated, and if so, is there no way to share that resource?

proofit404:
I guess this is possible with custom setup.py command. For example Sentry ship it's frontend this way.

fcurella:
@andrewgodwin the code in 'static' is compiled from the code in the 'node/src'. Specifically, it is transpiled (via babel) _and_ (unlike the published module) exported into a global 'channels' object so that is available to the browser.

Could we remove the file in static, and automate runnning 'npm run browserify' as part of the release process somehow?

fcurella:
@proofit404 thanks for the pointer, I'll look into how sentry does it

fcurella:
I've looked into how Sentry handles it, and we could compile at installation time.

Considering that we don't want to make nodejs a requirement for channels, I'm thinking maybe this should be handled as an extra. Something like 'pip install channels[javascript]'

proofit404:
We can run babel before making sdist and wheels. This way node.js will be necessary for contributing to channels itself.

fcurella:
I didn't realize you could do that :) I'll work on something on Monday

andrewgodwin:
Yes, having it be part of the release/build process is likely the best thing. I don't want to make people install node if they use a `git`-based install, though, so it needs to be optional in some fashion.

The alternative is keeping it as-is and putting a giant comment at the top of the file saying it's autogenerated, and how to regenerate it if you submit a PR or similar. We could probably then add a rule to the Makefile that just has releasing in it at the moment to let you build it easily.

fcurella:
>  I don't want to make people install node if they use a `git`-based install

I could simply check for the existence of `npm` and silently bail out (or raise a warning) if it's not found

andrewgodwin:
I think a warning would be fine, provided it was obvious enough that people who were 1) installing from source and 2) wanted to use staticfiles noticed it (of course, some people wouldn't notice a warning if it was fullscreen, but eh.)

I'd also be fine with just capturing the build step into the Makefile, though, and having a comment on the file saying not to edit it directly. That way it comes ready to run from wherever.

stunaz:
wouldn't it be better to have this in his separate package?

andrewgodwin:
It's hard to say. On one hand, it's nice to have a single solution that we maintain and that comes packed in (and it's not JavaScript-framework specific, as far as I can tell). On the other hand, yes, maybe it would be easier to maintain and release separately?

fcurella:
> maybe it would be easier to maintain and release separately?

I don't believe using a separate repo makes it easier to maintain. If anything, switching often between repos can become cumbersome.

However, I do think we should consider how we want to release the JS wrapper.

I would expect most of the work on the JS wrapper to be due to changes (most likely, new features) to channels. So the question becomes: do we want to 'enforce' the JS to stay up to date, or do we want to allow it to be a little behind?

I think framework-specific wrappers should be external -- given how fast the JS ecosystem moves.

But I also believe (although I can't know for sure) that the vanilla wrapper would be much more stable and won't need much maintaining, and most of the times the maintainer won't have to run `make build_assets`

andrewgodwin:
Right, as it's written right now I don't think much is ever going to change, and the `channels` package is specifically the "nice websocket bits" layer on top of Django, rather than all of the code (the hard stuff lives in the ASGI/Daphne packages).

I think I would want the `node` directory renaming though, as right now it's a bit mysterious and might imply we have node-backed server code or something. I'd prefer something like `js_client`.

fcurella:
@andrewgodwin I've renamed the directory to `js_client`

andrewgodwin:
Alright, looks good - ready to merge?

proofit404:
@fcurella thank you for all efforts you're putting into this issue.

Maybe we can write this module in the ES5 syntax. This way we can avoid compile assets step. So we can just put this module into `static` directory.

I wander if it make things much simpler? What do you think?

fcurella:
There are 2 steps involved in compile assets for `static`:

1. Transpiling: This converts the ES6 syntax to ES2015, so that most browsers can parse it
2. Bundling: This step basically concatenates all the `import ...`d (or `require()`d) files into the destination file.

Writing directly in ES2015 will spare us from transpiling, but we'll still have to bundle the module with its requirements.

proofit404:
The only dependency I see is reconnecting websocket.  We can check if it installed and fail the same way as we do in asgi_redis for cryptography of twisted imports.

fcurella:
I liked the idea of having a module on npm to act as 'baseline', upon which wrappers for specific js frameworks can be built. But I don't see how we could have both the module and the `static` script without bundling.

I also really liked the idea of providing users with a simple, ready-to-use javascript wrapper that they could just add to their pages. Unbundling `reconnecting-websocket` will mean more burden on the user side. We can make it as easy as copy and paste by writing documentation, but it will still mean one more step for users: 1. add `reconnecting-websocket` 2. Add `channels`

proofit404:
Yeh, maybe build simplicity don't worth it.

andrewgodwin:
@fcurella I agree with you on the build stuff here - ready for me to merge this in? I'd like to get a 1.1 release out with this in it soonish.

fcurella:
@andrewgodwin it's good to go. I've also incremented version on the package.json so it's ready to be published.

andrewgodwin:
Great, thanks. Here we go!

PR #555, opened 2017-03-10
Make the channel_session decorator work on methods as well

ErwinJunge:
I've modified the `channel_session` decorator so that it can be used on methods of class-based consumers as well.

### DIFF
--- a/channels/sessions.py
+++ b/channels/sessions.py
@@ -42,7 +42,7 @@ def channel_session(func):
     @functools.wraps(func)
     def inner(*args, **kwargs):
         message = None
-        for arg in args:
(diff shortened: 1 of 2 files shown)


proofit404:
Any tests for this improvement?

ErwinJunge:
@proofit404 Added tests

andrewgodwin:
The one change I would probably make to this is only checking the first two arguments; I'm worried that this is a little bit too permissive about where the message is allowed to be in the argument list. Otherwise, I like it.

ErwinJunge:
@andrewgodwin Done :)

andrewgodwin:
Great, thanks!

PR #564, opened 2017-03-22
Channel and http session from http

Coread:
expose the feature in issue #318 to generic websocket consumers.


andrewgodwin:
Nice, thanks!

PR #569, opened 2017-03-23
Allow runworker to be used without staticfiles

ErwinJunge:
While building a very minimal project that doesn't use the django staticfiles app at all, I ran into a problem where runworker tried to use the django staticfiles app and thus complained about my missing STATIC_URL setting. This PR fixes that by checking whether the project actually uses staticfiles before trying to handle requests for it.

### DIFF
--- a/channels/management/commands/runworker.py
+++ b/channels/management/commands/runworker.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+from django.apps import apps
 from django.conf import settings
 from django.core.management import BaseCommand, CommandError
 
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Ah right, we fixed this in `runserver` in https://github.com/django/channels/commit/b2842f1ef1d2365971268a351b39ed9741b539bd, but yes, `runworker` needs the patch too. I'll take it in once the build passes.

ErwinJunge:
Build fixed. Was a strange issue to debug, since it didn't trigger once when I ran tox locally. It turned out that the bindings from test_binding.py stuck around, making the tests not isolated enough. I fixed it by forcefully unregistering all bindings for the runworker tests, but this issue might need some more investigating.

andrewgodwin:
That seems correct to me, likely we should have had that in there before.

PR #572, opened 2017-03-24
Expose `WebSocketBridge.socket`

fcurella:
Made `WebSocketBridge._socket` public.

We can provide a convenient way to hook custom behavior (such as debugging) by making the underlying `ReconnectingWebSocket` instance publicly available.

Also removed the undocumented `onopen` constructor option.


andrewgodwin:
Seems like a good idea!

greut:
Do you mind pushing it to npm? Cheers,

andrewgodwin:
It'll go up with the next release - our PyPI and NPM releases are in sync. That should be tomorrow or the day after.

greut:
Thanks, I've monkeypatched it in the meantime `ws.connect; ws.socket = ws._socket`

PR #589, opened 2017-04-04
Improve docs.

proofit404:
A friend of mine had some troubles figuring out how to make `Models` chapter examples works. I think we should make this part more explicit.


andrewgodwin:
Looks good. Not sure why the build failed; I suspect it's a random failure as there were npm issues yesterday, so I'll pull it in and fix the master if it fails again.

PR #593, opened 2017-04-06
Added a decorator that checks the origin header of a websocket.connec…

### LINKED ISSUE #585: use of the CSRF cookie for authentification
Otherwise an user is already logged in, in all new connections even from other sites witch makes it attacking quiet easy.

dhepper:
…t message. This partially fixes #585

I created a new module at the top level because I wasn't sure where to put it. It should probably be moved to a more appropriate location.

For a working example, check out https://github.com/dhepper/channels-examples/tree/origin_check

### DIFF
--- a/channels/origin_validation.py
+++ b/channels/origin_validation.py
@@ -1,0 +1,80 @@
+from functools import update_wrapper
+
+from django.conf import settings
+from django.http.request import validate_host
+
+from .exceptions import DenyConnection
+
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
+
+
+
+
+class BaseOriginValidator(object):
+    """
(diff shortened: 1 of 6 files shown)


andrewgodwin:
I like the design of this, it's clean and looks good! Two things:

* Let's call it `channels/security/websocket.py`, which fits more with how we try to separate things a little.
* Are you able to include basic docs in the PR? Not everyone is happy writing docs, and if you can't, I can do it, it'll just take a bit longer :)

dhepper:
Thanks for your quick feedback!

* I've renamed the new module to `channels/security/websockets.py` (with a trailing s to match the existing modules)
* I've added some docs and extended the doc strings a bit.

andrewgodwin:
Thanks so much! Looks great, and merged in.

Issue #598, opened 2017-04-08
Delay server support for handling ChannelFull exception gracefully


sachinrekhi:
I'm planning on using delay server to schedule a task to generate an email at a certain time for all users. When the message becomes due, there will be a whole set of delayed messages that are sent to a specific channel for which I have a dedicated pool of back-end workers ready to handle.

However, glancing at the delay server runloop code, it didn't look like it handled the ChannelFull exception, so I'm concerned that when a whole set of delayed messages is sent to the workers, the channel could easily become full with the many now-due messages, resulting in many of those messages being dropped.

I would have expected the delay server to gracefully handle the ChannelFull exception and retry sending those messages to the channel at some near point in the future.

Is my understanding correct that these delayed messages could drop if the channel is full? And where do I find or set what the capacity limit is for my channel to potentially get around this issue in the near-term (using redis channel layer)?

Is what I'm suggesting a reasonable addition to delay server?

andrewgodwin:
Yes, that's a very reasonable addition to the delay server, it should re-queue them if it cannot send any. The capacity limits for all channels are set by the defaults in your ASGI backend and then overrideable by you - by default it is probably 100.

sachinrekhi:
Now implemented here: https://github.com/django/channels/pull/600

PR #600, opened 2017-04-09
Delay server will now requeue messages for later when gets ChannelFull exception

sachinrekhi:
When the delay server processes and sends delayed messages to a channel that is full, it will requeue those unsuccessfully sent messages to be sent after a delay.

This addresses https://github.com/django/channels/issues/598

Added a test case to ensure that when channel is full, the delayed message is in-fact requeued for later.

I set the requeue_delay as 1 second and it's not really user-configurable. Let me know if you have an alternative recommendation on a default requeue_delay.


andrewgodwin:
This is good enough for now, I think - we can come back and make the requeue configurable if it proves to be a problem.

PR #603, opened 2017-04-12
Provide a way to disable static files serving in the live server test case.

proofit404:
As we discussed in the #497 user should be able to disable static files serving in the live server tests.


andrewgodwin:
:+1:

PR #604, opened 2017-04-12
Provide TEST_CONFIG alias.

proofit404:
As we discussed in the #497 we need test aliases to prevent integration test overlap with running dev environment.


andrewgodwin:
Looks good! Just needs to be rebased against the other one I just merged, I think.

proofit404:
Done.

Issue #601, opened 2017-04-11
Test message has no 'method' or 'client' keys


gyermolenko:
Hi.

When I submit message to channel via HttpClient, like so
```client.send_and_consume('websocket.connect', {'value': 'my_value'}, path=url)```

in consumer I get message (channels.message.Message object) without 'method' or 'client' keys (and therefore can't transform it with AsgiRequest(msg) and such).

Could you clarify if it is a bug?

ps 
channels==1.1.2 
asgi_redis==1.2.1

andrewgodwin:
I believe it may be a bug - @proofit404, what do you think?

proofit404:
First of all code above gives me `TypeError` which is obvious because `send_and_consume` has different signature.

I don't think this is a bug. Since Daphne construct message content itself, we force users to copy this internal behavior in tests. Add right `method` and `client` keys to the content dict is good example.

I think we can add helper method like `client.http_send_and_consume` which will provide right content. But I'm against `send_and_consume` modification because it should not be coupled to the HTTP protocol.

andrewgodwin:
Yeah, I've seen a couple of bug reports about this now but it doesn't make sense that it would work like this - last time I recommended to use the plain `Client`. I think the problem is that the testing docs make use of it in this way here: https://github.com/django/channels/blame/master/docs/testing.rst#L160

proofit404:
It's nice to have http helper. If `method` key is required to handle this message (I guess it is so, because it goes into django view machinery), we need provide `method` key. Maybe even make it consumer method keyword argument.

But I'm not so certain about `client` key. It probably required by some http middleware, but is it necessary in general?

Krukov:
I know, that is strange, but the `HttpClient` is for testing webscoket consumers (not http). It seems that this introduces misunderstandings.

proofit404:
Does `HttpClient` useful only for websockets? Maybe we can have both http and ws test clients? Or have one "web" client with send http and send ws methods.

gyermolenko:
> First of all code above gives me TypeError which is obvious because send_and_consume has different signature.
@proofit404 hi. Could you please give me some advice regarding the situation?
How should I approach testing consumers, what is the correct signature for send_and_consume? Should I use this method at all? Should I wait for updates in documentation / next channels release?
Alas, for the moment I'm just confused with your conversation...
Thank you in advance.

Krukov:
[...]

Issue #586, opened 2017-03-31
Demultiplexer testing with enforcing order 


cjds:
Channels 1.1.1

I'm tryign to do a test with a Demultiplexer, and it complains about enforcing ordering. I have the flag set, but do I have to add an order value to every message I send?
```
class ROSChannelTestCase(ChannelTestCase):
    """
    ROS Channel Test Case
    Test if we can do things
    """

    def test_ros_channel(self):
        client = HttpClient()
        # user = User.objects.create_user(username='test', email='test@test.com', password='123456')
        # client.login(username='test', password='123456')
        with apply_routes([Demultiplexer.as_route(path='/new/')]):
            client.send_and_consume(u'websocket.connect', path='/new/')
            self.assertIsNone(client.receive())
```

>   File "/home/cjds/development/robot_configuration_interface/robot_configuration_interface/tests/unit/test_channels.py", line 35, in test_ros_channel
>     client.send_and_consume(u'websocket.connect', path='/new/')
>   File "/usr/local/lib/python2.7/dist-packages/channels/test/http.py", line 95, in send_and_consume
>     return self.consume(channel, fail_on_none=fail_on_none, check_accept=check_accept)
>   File "/usr/local/lib/python2.7/dist-packages/channels/test/http.py", line 98, in consume
>     result = super(HttpClient, self).consume(channel, fail_on_none=fail_on_none)
>   File "/usr/local/lib/python2.7/dist-packages/channels/test/base.py", line 136, in consume
>     return consumer(message, **kwargs)
>   File "/usr/local/lib/python2.7/dist-packages/channels/generic/base.py", line 32, in __init__
>     self.dispatch(message, **kwargs)
>   File "/usr/local/lib/python2.7/dist-packages/channels/generic/base.py", line 70, in dispatch
>     return self.get_handler(message, **kwargs)(message, **kwargs)
>   File "/usr/local/lib/python2.7/dist-packages/channels/sessions.py", line 78, in inner
>     return func(*args, **kwargs)
>   File "/usr/local/lib/python2.7/dist-packages/channels/auth.py", line 42, in inner
>     return func(message, *args, **kwargs)
>   File "/usr/local/lib/python2.7/dist-packages/channels/generic/websockets.py", line 68, in raw_connect
>     self.connect(message, **kwargs)
>   File "/usr/local/lib/python2.7/dist-packages/channels/generic/websockets.py", line 275, in connect
>     consumer(message, **kwargs)
[...]

PR #609, opened 2017-04-16
Patch rename http client

Krukov:
Ref #601 #586


andrewgodwin:
I don't think we can just remove it like this entirely - I'd rather make it capable of both (and it seems to be mostly there)?

Krukov:
Yes, maybe removing is wrong way, but I think that separation is necessary. Http client should have more http friendly methods, like get, post ..... I think they do not need to be mixed.

andrewgodwin:
OK, let's do this for now then, I need to sit down and work out the testing stuff properly.

ftxrc:
It would be great if this was in the docs; if I have time I might PR it.

andrewgodwin:
That would be most welcome.

PR #615, opened 2017-04-19
Allow to start/stop server and worker in ChannelLiveServerTestCase

jricardo27:
This would allow to restart worker when required, e.g.
Using `Freezegun` to mock the time doesn't have effect unless the
worker is restarted.

### DIFF
--- a/channels/test/liveserver.py
+++ b/channels/test/liveserver.py
@@ -170,6 +170,45 @@ class ChannelLiveServerTestCase(TransactionTestCase):
 
         return 'ws://%s:%s' % (self.host, self._port_storage.value)
 
+    def start_server_process(self):
+
+        server_ready = multiprocessing.Event()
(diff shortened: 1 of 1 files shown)


proofit404:
You can mock time in the whole class. Also I don't think that mocking should be considered as a reliable technique in the integration tests. Your tests will fail anyway if multiprocessing mechanism have changed from fork to spawn or spawnserver. Also this will not work at all on platform without fork.

proofit404:
The proper way of mocking in this way, if you **really** need this, override `run` method in the subprocess and pass it to the test [this](https://github.com/django/channels/blob/master/channels/test/liveserver.py#L158) way.

andrewgodwin:
Yeah, I'm going to say that we can't support in-memory mucking around with stuff with the live test case and close this. It's not really feasible since they have to live in different processes; you should treat the test case more like a full acceptance test.

PR #625, opened 2017-04-28
Two minor spec changes

maikhoepfel:
This pull request implements two minor spec updates that came up during my work on https://github.com/django/daphne/pull/111.


andrewgodwin:
:+1:

PR #627, opened 2017-05-03
add in the file binding/websockets.py the condition

debianitram:


proofit404:
`self.fields` can be a tuple for example. This way condition statement wont match, since you removed type cast.

debianitram:
Sorry, but I'm not agree. If the fields property is a tuple with the element ('__all__'), python considers as a string '__all__'.

Execute: type(('__all__'))

Sorry my bad english. Gato! :P

andrewgodwin:
I agree with Artem. In Python, `("__all__",)` is not equal to `"__all__"`.

proofit404:
And `"__all__"` is not equal to `"__all__",`. I don't like this part of python syntax for sure.

debianitram:
Ok, ("__all__"**,** ) is not equal to "__all__", thats true. 
What I want to say, is that if you considers the fields property with the value ('__all__', ) don't have any sence. 
You should specify in the documentation how to use this property adequately.
If I set fields property with value '__all__', when I use the method update not change data in database.

Sorry my english!

debianitram:
__all__ *

debianitram:
`__all__` *

proofit404:
I don't see why `("__all__",)` can be illegal use case.  Django forms behave the same way. So at least we should keep things convenient.

debianitram:
Ok, then:
`self.fields in ('__all__', ['__all__'], ('__all__',))` ¿?

andrewgodwin:
Binding is now gone, so closing this, sorry we never got to merging it!

Issue #651, opened 2017-05-30
HTTP/WS send decode error: 'dict' object has no attribute 'encode'


mssever:
It appears that `Group('channel').send()` is trying to call `encode()` on `dict` objects, rather than the inside `str`s. `runserver` reports the following error:

    ERROR - server - HTTP/WS send decode error: 'dict' object has no attribute 'encode'

Sending an empty `dict` doesn't error out, but an actual (JSON-serializable) `dict` causes this error. To try to track it down, I subclassed `dict` to add my own `encode()` method:

    class MyDict(dict):
        def __init__(self, **kwargs):
            for k, v in kwargs.items():
                self[k] = v
    
        def encode(self, *args, **kwargs):
            print('args: {}\tkwargs: {}'.format(args, kwargs))
            return self

The print statement shows that it's getting called as `the_dict.encode('utf-8')`. Is it trying to call the method on some part of the dict? All the docs say is that `send()` expects a dict, and when I tried passing in a `str`, I got an exception to the same effect. Yet, it's calling a method which doesn't exist on a normal dict.

My environment:

- Ubuntu 16.04
- Python 3.5.2
- Software versions:
    - asgiref (1.1.2)
    - channels (1.1.3)
    - daphne (1.2.0)
    - Django (1.11.1)
    - Twisted (17.1.0)
    - ASGI backend: Just the default in-process one for now
- What you expected to happen vs. what actually happened: I expected to be able to exchange messages without errors.
- How you're running Channels: runserver
- Console logs and full tracebacks of any errors: See above. No tracebacks were generated, unfortunately.

andrewgodwin:
So this is happening somewhere in Daphne, and I suspect it is that one of the keys in the dictionary is a dictionary when it is not expected to be. Could you give an example dump of the message content you are trying to send that is causing the error?

mssever:
You're right that one of the keys is a dictionary. Here's a sample message:

    {'text': {'sync': True, 'type': 'change_slide', 'data': 1}}

It's that way because I get another error if I send just the data. If I send `{'type': 'change_slide', 'sync': True, 'data': 1}`, I get:

>  ERROR - server - HTTP/WS send decode error: Got invalid WebSocket reply message on daphne.response.SYyQvGyXkF!KxwjKNQMqk - contains unknown keys {'data', 'type', 'sync'} (looking for either {'accept', 'text', 'bytes', 'close'})

[...]

PR #653, opened 2017-05-31
Update the docs to provide info on how to use Group.send()

mssever:
Discussion in bug #651 revealed that the docs didn't explain how to use `Group.send()`.  This pull request attempts to address this issue.

Rationale
=======
I was initially referred to [the specification](http://channels.readthedocs.io/en/latest/asgi/www.html). The problem is that specs are for those who are implementing the relevant protocol. They aren't for API users. A user of the API shouldn't have to know anything of the internals unless they want to alter the internals. An API reference should provide all of the information necessary to correctly use it.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -88,7 +88,17 @@ They have the following attributes:
   representing the underlying channel layer to send messages on.
 
 * ``send(content)``: Sends the ``dict`` provided as *content* to all
-  members of the group.
+  members of the group. The ``dict`` keys are as follows:
+
+  * ``bytes``: Byte string of frame content, if in bytes mode, or ``None``.
+  * ``text``: Unicode string of frame content, if in text mode, or ``None``.
+  * ``close``: Boolean indicating if the connection should be closed after data
+    is sent, if any. Alternatively, a positive integer specifying the response
+    code. The response code will be 1000 if you pass ``True``. Optional, default
+    ``False``.
+  * ``accept``: Boolean saying if the connection should be accepted without
+    sending a frame if it is in the handshake phase. Optional.
+  * Note: A maximum of one of bytes or text may be provided.
 
 * ``add(channel)``: Adds the given channel (as either a :ref:`Channel <ref-channel>`
   object or a unicode string name) to the group. If the channel is already in


andrewgodwin:
Unfortunately I can't take this - group send can be for anything, not just WebSockets, though that is obviously the normal case for most Channels users. Is there somewhere else you think it might be better suited that isn't generic reference docs for the send API?

mssever:
If not there, where? Imagine the following scenario:

1. A user (me in this case) is new to Channels and has no prior knowledge of the project.
2. The user reads the tutorial and gets a basic route set up. Setting up the JavaScript side is easy.
3. The user doesn't believe that he/she needs the `text` key that the examples use, believing it to just be part of the example, rather than part of the API, so he/she uses keys which are relevant to the project at hand.
4. `runserver` throws an error, complaining about unexpected keys.
5. The user goes to the logical place to learn how to use `send()`, the API docs.
6. The docs say that `send()` expects a `dict`. The logical conclusion is that this means _any `dict`_ and that that the docs are wrong. The docs don't provide any information to help the user learn how to use the API, which means in their current state they might as well not exist.
7. The user Googles the error message and comes up with nothing which appears relevant. A few unrelated bug reports.
8. The user wants to know what the expected keys mean, but there's no information to be had, and no way to find out.
9. After trial and error, the user figures out that the appropriate key is `text`, which expects a string, and that all the other keys are irrelevant. (Since other docs mentioned that the API would call JSON on supplied dictionaries, the fact that `text` expects a string and not a dictionary is not obvious.)

If what `send()` expects changes depending on something, then whatever that is should be explained in the API docs, and links should be given to make it possible for users to discover what it expects in their case. I can't submit such a patch because I haven't got a clue what the answer is. All I know is that, if you read the API docs and don't learn how to use the API, the docs aren't doing their job.

It seems like this project is only for those who are already familiar with it. For those who are unfamiliar with it, good luck. That, or being generic is emphasized so much that real-world usefulness is seen as of lesser importance.

andrewgodwin:
Well, ideally, your list would end at:

2. The user reads the tutorial and it makes clear that `text` is a WebSocket-specific thing

Or.

3. The user reads the code examples repo where it's made clear how the encoding works

I'm not saying we don't need to improve the docs, I'm saying that the specific location you PR'd against is generic. I wouldn't be averse to putting a `note` admonition there that calls out that you need to consider the sending format and linking out to an explainer section elsewhere.

PR #654, opened 2017-05-31
Added shortcut to simplify the common case of sending JSON text on a channel

mssever:
While the Channels' send API is generic, it can be helpful to have some shortcut classes available for common uses, to simplify the use of Channels. This pull request adds a `JSON_Dict` class (a subclass of `dict`) which makes it easier to send JSON data over a channel, such as a WebSocket.

Here's an example of how it might be used:

    from channels import Group
    from channels.shortcuts import JSON_Dict
    
    def send(**kwargs):
        Group('somegroup').send(JSON_Dict(**kwargs))

This improves encapsulation by not requiring API users to know or care much about the send protocol--at least if this shortcut matches their use case.

### DIFF
--- a/channels/shortcuts.py
+++ b/channels/shortcuts.py
@@ -33,7 +33,7 @@ class JSON_Dict(dict):
 
     def update(self, *args, **kwargs):
         __doc__ = self._data.update.__doc__
-        [__doc__] # Silence test complaint about __doc__ being assigned but not used.
+        [__doc__]  # Silence test complaint about __doc__ being assigned but not used.
         self._data.update(*args, **kwargs)
         self._update_text()
 
@@ -52,7 +52,7 @@ class JSON_Dict(dict):
             raise ValueError("The key name 'text' is reserved.")
         del self._data[key]
         self._update_text()
-    
+
     def __repr__(self):
         return 'JSON_Dict({})'.format(repr(self._data))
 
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I'm not a fan of this, so I'm going to say no - I don't see how

```
from channels.shortcuts import JSON_Dict
Group('somegroup').send(JSON_Dict(**kwargs))
```

is much better than
```
import json
Group('somegroup').send(text=json.dumps(kwargs))
```
I don't believe in adding abstractions in a case like this when educating people about the underlying protocol is about as easy and better in the long run for designing good software.

mssever:
Why should it be necessary to know the underlying protocol? To use Django with HTTP I don't need to know the HTTP protocol. If I have more advanced needs, sure, but normally I can just use Django's API. This is even more true when discovering the underlying protocol is extremely difficult (as #653 attempts to address).

andrewgodwin:
Django is much more tied to HTTP than Channels is to WebSockets; in particular, I have IRC and Slack protocol servers that would use code that looks very similar apart from a different response format.

Channels already has wrapping around JSON handling and WebSockets, too, in the `JsonWebsocketConsumer`, which overrides its own `self.send` as you describe: http://channels.readthedocs.io/en/latest/generics.html#websockets - this is the high-level interface we intend for people to use if they are doing JSON-WebSocket.

Issue #656, opened 2017-05-31
Don't allow enforce_ordering on a multiplexed sub-consumer


millar:
We're having problems using `string_ordering = True` on multiplexed consumers:

If you use >1 consumer, the connect event message will be sent to each consumer registered on the demultiplexer. This breaks ordering since the `order` value in this message (as copied to all consumers) is `0`.

The `__channels_next_order` value will be incremented after the first consumer is finished, then the next consumer tries to handle the connect event but it can't since the message `order` value is 0 and `__channels_next_order` on the session is 1.

Subsequent messages are handled OK but if, for example, data should be stored in `channel_session` by any of the stuck connect event handlers this will cause problems should the messages rely on it (as they do for us).

I have produced a workaround this by tweaking the `enforce_ordering` decorator but I'm not convinced it's the best approach (I will submit this as a PR nonetheless).

andrewgodwin:
Ah yes, the interaction of the demultiplexer with `enforce_ordering` is problematic. I would say that the order should be enforced at the demultiplexer and we should instead raise an error if you try to demultiplex into an enforce_ordering consumer - what do you think?

millar:
That'd definitely work to solve the problem – and an exception would probably be useful to stop people making the same mistake!

The reason we decided to do it on a per demultiplexed channel basis was to reduce the performance hit on channels where order does not matter. I imagine the overhead of fixing that type of implementation would outweigh the practical performance benefits?

andrewgodwin:
Right, doing it once per multiplexed channel is exactly as bad as once on one of the demultiplexed sub-channels, and always better if you're doing it more than one sub-channel, so it's the only real way these things should be done.

andrewgodwin:
`enforce_ordering` is gone! So this bug is getting closed.

PR #657, opened 2017-05-31
Fix strict ordering with multiplexed consumers

millar:
My solution to #656. I have added a test that fails without the changes to the `enforce_ordering` decorator.

**This PR does not solve the equivalent problem with demultiplexer `disconnect` events, which means we probably need a more robust solution! Any ideas?**


andrewgodwin:
Ordering is now gone in Channels 2.0 as it's all in process, so closing this. Sorry we never got back to you!

PR #659, opened 2017-06-05
Fixes #612 - Log channels to "django.channels.server" instead of stderr

### LINKED ISSUE #612: runserver should log to "django.channels.server" instead of stderr

renatooliveira:
I updated django's DEFAULT_LOGGING dictionary to add channels options.

There is a small code duplicate from django configure_logging. If I don't do this, I remove the user option to create their own LOGGING in settings.

I'm still not sure how I should test this, any guidance?

### DIFF
--- a/channels/apps.py
+++ b/channels/apps.py
@@ -1,6 +1,8 @@
 from django.apps import AppConfig
+from django.conf import settings
 
 from .binding.base import BindingMetaclass
+from .log import configure_logging
 from .package_checks import check_all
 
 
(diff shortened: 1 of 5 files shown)


andrewgodwin:
This seems fine to me given what it's replacing. Did you mean to close this?

renatooliveira:
Yes, tests failed. I found an error. 

Current status is: DEFAULT_LOGGING['formatters'] was added in 1.10, trying to figure out a way to keep compatibility with all versions

Issue #677, opened 2017-06-22
Docs: "we’ll cover this later." ... where=


guettli:
On the page "Channel Concepts", chapter "How do we use channels?", you say _we’ll cover this later_

https://channels.readthedocs.io/en/stable/concepts.html#how-do-we-use-channels

For me this is actually the most important detail.

I am unsure which section you mean with "later".

It would be very nice if you could make this a link to the relevant section.

Thank you very much.

ftxrc:
I'm sure it means Daphne. specifically this section: http://channels.readthedocs.io/en/stable/deploying.html?highlight=daphne#run-interface-servers
I'll try to commit an edit.

PR #678, opened 2017-06-22
Link to interface server explanation in concepts

ftxrc:
The docs did not link to the interface server explanation. Relates to issue #677


guettli:
Andrew and Sergio: Thank you for loving and fixing docs.

PR #681, opened 2017-06-23
Improve docs consistency on query strings

ftxrc:
Change to room_name parameter and erase references to query strings throughout the documentation tutorial.

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -364,7 +364,7 @@ Persisting Data
 Echoing messages is a nice simple example, but it's ignoring the real
 need for a system like this - persistent state for connections.
 Let's consider a basic chat site where a user requests a chat room upon initial
-connection, as part of the query string (e.g. ``wss://host/websocket?room=abc``).
+connection, as part of the URL path (e.g. ``wss://host/rooms/room-name``).
 
 The ``reply_channel`` attribute you've seen before is our unique pointer to the
 open WebSocket - because it varies between different clients, it's how we can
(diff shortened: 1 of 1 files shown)


andrewgodwin:
Yup, that's better, thanks!

dsanders11:
This has already been merged, but @ftxrc did ask for a thorough review.

While it removed the query string inconsistency, now the documentation doesn't explain where `room_name` is coming from, and the example code is wrong.

The `routing.py` example for the `Persisting Data` section does not extract the room name from the URL path, so the example code under `Persisting Data` is incomplete. This could be fixed by updating the `routing.py` example, but then it would be using a feature that isn't explained until later in the `Routing` section (not the end of the world). In order to match with the code it would also need to have the URL path in all three WS message types, unlike the later `Routing` section which only pulls `room_name` for `ws_add`.

My 2 cents.

ftxrc:
@dsanders11 I'll definitely explain it a bit more in a bit. I did add it to the routing.py, just the main one however; that one seems to be called in the main rooms.py where the websocket calls are routed.

```
chat_routing = [
    route("websocket.connect", chat_connect, path=r"^/(?P<room_name>[a-zA-Z0-9_]+)/$"),
    route("websocket.disconnect", chat_disconnect),
]
```
However, reading it again, some things in it are slightly wrong; I'll fix them in about 30 mins.

dsanders11:
> I did add it to the routing.py, just the main one however; that one seems to be called in the main rooms.py where the websocket calls are routed.

But you missed the one [in `Persisting Data`](http://channels.readthedocs.io/en/stable/getting-started.html#persisting-data) right after the example code. Right after the code you updated to add the `room_name` argument, it says "Update routing.py as well:" and that code does not include the URL path for `room_name`. Lines ~410-430.

ftxrc:
@dsanders11 I'll see; I'll do this once I have lunch. Forgot about yesterday haha, will take a look and fix.

PR #688, opened 2017-06-28
Upgrade package.json to NPM v5

fcurella:
I've update the `package.json` to the latest npm changes.

something I'm not clear about is if we should include `package-lock.json` or `shrinkwrap.json`. If we include `shrinkwrap.json`, it will need to be updated for every release, and I'd like to not have any additional step to the release process


proofit404:
If we include lock file, we will improve "reproducability". But I'm not sure if it worth it.

andrewgodwin:
Yeah, let's leave the locking out for now.

PR #691, opened 2017-06-29
fixed: some device will get error code [1006] when handshake if proto…

bmcool:
fixed: some mobile device will get error code [1006] when handshake if protocols is undefined

ex: HTC M8, SAMSUNG Galaxy S8


andrewgodwin:
Since this is a JS change, can you also run the rebuild process for the bundled code? I like to keep the main repo up to date.

PR #692, opened 2017-06-29
Add note about Live Server Test Case and SQLite

magoogli:
By default Django will use SQLite as an in memory database when running tests. This will not work with the Live Server Test Case. Added some notes on how to configure the database so that it is not run in memory.


proofit404:
Great! Thanks.

PR #693, opened 2017-06-29
Fix #612 - Port channels log from stderr to logging framework

### LINKED ISSUE #612: runserver should log to "django.channels.server" instead of stderr

renatooliveira:
#612 

`channels/log.py` is basically a copy of django log itself.

Some decisions I made here that I don't know if they're right:

1 - I moved the logger object away from the class because it was being called before instantiation
2 - I removed formatters from DEFAULT_LOGGING due to backwards compatibility (It was only added on 1.10)

I'm also not sure if the tests are right, and would like some guidance on this.

Looking forward to hear your feedback!

Thanks!


renatooliveira:
I just saw that the checks failed, I'll fix it.

renatooliveira:
hey @andrewgodwin is there anything I can do to help with this PR?

proofit404:
I'll review it during this week. Thanks for your contribution!

renatooliveira:
Thanks @proofit404 I'll do the changes this week and let you know

channels/log.py
```diff
@@ -1,31 +1,88 @@
...
+        'django': {
+            'handlers': ['console', 'mail_admins'],
+            'level': 'INFO',
+        },
+        'django.server': {
+            'handlers': ['django.server'],
+            'level': 'INFO',
+            'propagate': False,
+        },
+        'django.channels.server': {
```
renatooliveira: Hi @proofit404 what should I use here? I used what was proposed on the github issue.

Thanks

proofit404: I'll reread related issue and leave a more helpful comment.

proofit404:
Since channels package isn't part of the Django code base, I don't think its good idea to use `django.channels.server` as the logger name.

Could you answer what behavior we expect, if someone run development server with `--use-asgi=false`?

Also, can you write documentation chapter on configuration logging with changes you've done?

renatooliveira:
Hey @proofit404, I changed from `django.channels.server` to `channels.server`.

Regarding `--use-asgi=false`, you mean `--noasgi`, right? I'm not sure which behavior I should expect on this. I tested against master and my branch and both worked the same way.

I'll start writing the docs for this

jjlorenzo:
any updates?

proofit404:
I'll return to this PR at the weekend.

proofit404:
I've tried this PR. Looks like it working. But I need more time to review it intently. Sorry for the delay.

slampenny:
Really looking forward to this merge. It's been very hard to debug without access to the log.

amir-hadi:
One side question: Why would you set propagate to False on the loggers? The old implementation breaks up my logging config: It sets propagate = True, even if I configure it on my own in the global LOGGING (dictConfig) config. I believe this change would respect the global configuration of the logging better, so if I set propagate = True on the channels logger, it will stay like that.

The reason why I am asking: From my understanding when you set propagate to False the log messages don't get forwarded to the root logger and so global handlers are not invoked. I have a SentryHandler on the root logger, which means every error or exception gets reported to Sentry (which is desired). So why would one set propagate = False as default?

andrewgodwin:
This PR is now super mismatched to the current release so I'm going to close it, but I would still like to see logging improved somewhat (though I did some improvements in Channels 2 anyway)

PR #705, opened 2017-07-19
Channels is comprised of six packages

koole:


proofit404:
Thanks! `asgi_rabbitmq` is not official yet. Nevermind.

koole:
Ah okay! I just noticed it has been there for 3 months already but the number was never updated.

PR #710, opened 2017-07-27
WIP Worker decoupled from channels.

proofit404:
In current implementation channels worker do constant polling of the channel layer. Some channel layers can benefit from worker based on underlying technology internals. I want to give an ability to the layer to substitute default behavior with its own worker.

This PR allow public tracking for related issue

- [ ] Main channel layers PoC implementations
  - [ ] RabbitMQ
  - [ ] Redis
  - [ ] IPC
- [ ] Generic Implementation
  - [ ] asgiref worker 
  - [ ] well defined API
  - [ ] conformance tests
  - [ ] spec documentation
  - [ ] inmemory worker will call this API directly
- [ ] Uvicorn integration
- [ ] Channels code cleanup
  - [ ] Remove worker module

### DIFF
--- a/channels/management/commands/runworker.py
+++ b/channels/management/commands/runworker.py
@@ -8,7 +8,7 @@ from channels import DEFAULT_CHANNEL_LAYER, channel_layers
 from channels.log import setup_logger
 from channels.signals import worker_process_ready
 from channels.staticfiles import StaticFilesConsumer
-from channels.worker import Worker, WorkerGroup
+from channels.utils import apply_channel_filters
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


auvipy:
Genecir Implementation -> generic?

proofit404:
Yep, you're right. Thanks.

andrewgodwin:
I won't be able to look into this properly until I get back from holiday, but my intention was to build this into asgiref - I already started with https://github.com/django/asgiref/blob/master/asgiref/worker.py

andrewgodwin:
Also, given that channels 2's models would not require the worker class for the general case, I don't want to put too much work in before we know what's happening there :)

proofit404:
I think the best way we can implement it for Channels 1.0 is an installed app with separate management command.

Issue #748, opened 2017-09-22
origin header not detected by allowed hosts decorator; error in channels/security/websockets.py:41


japrogramer:
#747 So even tho I explicitly set the origin headers the decorator doesn't see it
```

(Pdb) ll                                                                                                                                                                                                                                                    
130         def test_added(self):                                                                                                                                                                                                                           
131             """Test if call adds if"""                                                                                                                                                                                                                  
132             client = WSClient()                                                                                                                                                                                                                         
133   ->        client.set_header('origin', settings.ALLOWED_HOSTS[0])                                                                                                                                                                                      
134             query = '''                                                                                                                                                                                                                                 
135                 mutation Delete ($uuid: String!){                                                                                                                                                                                                       
136                 deleteProduct (uuid: $uuid){                                                                                                                                                                                                            
137                     status                                                                                                                                                                                                                              
138                     }                                                                                                                                                                                                                                   
139                 }                                                                                                                                                                                                                                       
140                 '''                                                                                                                                                                                                                                     
141             foovar = {'uuid': self.product1.uuid}                                                                                                                                                                                                       
142                                                                                                                                                                                                                                                         
143             content = {'query': query, 'foovar': foovar}                                                                                                                                                                                                
144                                                                                                                                                                                                                                                         
145             with apply_routes([                                                                                                                                                                                                                         
146                     route('websocket.connect', ws_GQL_connect, path=r"^/gql"),                                                                                                                                                                          
147                     route('websocket.receive', ws_GQLData, path=r"^/gql")]):                                                                                                                                                                            
148                                                                                                                                                                                                                                                         
149                 import pdb; pdb.set_trace()                                                                                                                                                                                                             
150  ->             client.send_and_consume('websocket.connect', path='/gql')                                                                                                                                                                               
151                 client.send_and_consume('websocket.receive', path='/gql', text=str(content))                              
152                                                            
153                 # message = self.get_next_message("input", require=True)                                                  
154                 # consumer(message)                        
155                 result = self.get_next_message(message.reply_channel.name, require=True)                                  
(Pdb) client.headers                                           
{'origin': 'test.com', 'cookie': b''}    
```
line 151 has a consumer with the decorator that fails with
```
Traceback (most recent call last):                             
  File "/usr/local/lib/python3.6/site-packages/channels/security/websockets.py", line 55, in get_origin                       
    header = self.get_header(message, b'origin')[0]            
[...]

PR #749, opened 2017-09-22
Fix that bug on issue #748

japrogramer:
Details of bug #748 
Fix is straight forward.


andrewgodwin:
This code is correct - hence the failing tests when you changed it. The problem is in the test client.

PR #756, opened 2017-09-28
Fix typos in "Getting Started" documentation text and example

tomka:
This is a very small change to the "Getting Started" documentation page. It fixes two minor issues I came across when reading through the document. A period character is added and a syntax error in one of the examples is fixed.


proofit404:
Thanks!

PR #765, opened 2017-10-17
preventing attribute error

freakypie:
Since `messages` is being deleted without being checked it is possible to generate a `AttributeError: messages`.

The for-loop above the potential error spot provides a default so it skips this potential


andrewgodwin:
I think in this case it would be safer to just catch the error and squash it in case of race conditions?

freakypie:
Could a message be added between the last send and the attribute removal?

andrewgodwin:
This class is now no longer in Channels, so closing. Sorry I didn't get back to you before.

Issue #650, opened 2017-05-29
Implement graceful exit in WorkerGroup


andrewgodwin:
It needs the ability to read the `in_job` status of all subthreads and set the `termed` flag so they all end up terminated or outside of a job before they quit. Follows on from #640.

sourabhtk37:
Anyone one working on this? @andrewgodwin 
Could you guide me? I'm new to the project.

andrewgodwin:
Nobody is working on this as far as I know - the way of implementing it is in the description I wrote (if you read through the worker code you'll see those variables in there, but you'll need to read though and understand it before fixing).

I wish I had time to help mentor you on this, but I'm really busy right now :(

andrewgodwin:
WorkerGroup is gone, and we did this anyway, so closing!

PR #790, opened 2017-11-21
Implement graceful stop for WorkerGroup

vshlapakov:
Addressing #650:
- added a new option to keep current behavior when it's appropriate
- added a helper to wait for all running workers in the group
- using it with SIGTERM handler and in the end of `run()` procedure

The cmdline option can be dropped if we agree that it should be default behaviour.

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -1,10 +1,13 @@
 from __future__ import unicode_literals
 
+import time
+import threading
+
 from channels import DEFAULT_CHANNEL_LAYER, Channel, route
 from channels.asgi import channel_layers
 from channels.exceptions import ConsumeLater
 from channels.test import ChannelTestCase
-from channels.worker import Worker
+from channels.worker import Worker, WorkerGroup
 
 try:
     from unittest import mock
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Patch looks good, though I think the commandline argument should default to True (but leave it in case people really want to turn it off).

Also, just so you're aware, Channels 2 is just around the corner and doesn't have this worker code in it, so your patch will only live on in the 1.x branch once that change happens!

vshlapakov:
@andrewgodwin Agree, inverted the option.

> Channels 2 is just around the corner and doesn't have this worker code in it

Thanks for the note, as of now I'm totally fine about it.

vshlapakov:
@andrewgodwin Added a simple test, polished and squashed, ready for review.

tests/test_worker.py
```diff
@@ -93,3 +96,36 @@ def test_normal_run(self):
...
+        def callback(channel, message):
+            callback_is_running.set()
+            # emulate some delay to validate graceful stop
+            time.sleep(0.1)
+            callback_is_running.clear()
+
+        consumer = mock.Mock()
+        Channel('test').send({'test': 'test'}, immediately=True)
+        channel_layer = channel_layers[DEFAULT_CHANNEL_LAYER]
+        channel_layer.router.add_route(route('test', consumer))
```
proofit404: This route pollute global state an doesn't cleanup it properly.

tests/test_worker.py
```diff
@@ -93,3 +96,36 @@ def test_normal_run(self):
         worker.run()
         self.assertEqual(consumer.call_count, 1)
         self.assertEqual(channel_layer.send.call_count, 0)
+
+
+class WorkerGroupTests(ChannelTestCase):
+
+    def test_graceful_stop(self):
+
```
proofit404: Add doc string what this test is actually do.

vshlapakov:
Thanks a ton @proofit404 :+1:

tests/test_worker.py
vshlapakov: Covered in https://github.com/django/channels/pull/792/commits/713e972bfecb6534c177169c498888ae8322bd0b.

tests/test_worker.py
vshlapakov: Fair, but I was searching for smth like `.remove_route` and found nothing, so the only way is to drop a route from `router.root.routing` manually and update channels names like it's done [here](https://github.com/django/channels/blob/master/channels/routing.py#L33-L34), right?

proofit404: Or you can create channel layer instance manually with class constructor and use it.

vshlapakov: Adressed the issue with https://github.com/django/channels/pull/792/files#diff-3ba621217225120eff2c061666b8043cR103, let me know if you like it or not.

PR #792, opened 2017-11-29
Improve graceful stop logic for WorkerGroup

vshlapakov:
This PR is a follow-up to #790 to handle an important case when a worker group gets a termination signal while there's a message being processed by main thread: as of now it may lead to an infinite loop and rejection to stop because [the call](https://github.com/django/channels/blob/master/channels/worker.py#L176) blocks the main thread (`while self.in_job or ..`).

A proposed work-around is to raise an exception from a signal handler (if main thread is awaiting for next message, so `self.in_job is False`), catch it within main thread and gracefully wait for sub-worker threads. Otherwise if main thread is processing a message right now, it will finish the processing step, exit from the main loop and end in a similar way.

@andrewgodwin @proofit404 Let me know what do you think, my deep apologises I didn't spot it before.

### DIFF
--- a/channels/worker.py
+++ b/channels/worker.py
@@ -16,6 +16,14 @@ from .utils import name_that_thing
 logger = logging.getLogger('django.channels')
 
 
+class StopWorkerGroupLoop(Exception):
+    """The exception is used to break worker group main thread loop from
+    SIGTERM/SIGINT handler when there is no job in process. If there is
+    a job in process, the loop will end by itself after processing it due
+    to self.termed flag and manual interruption is not needed.
+    """
+
+
 class Worker(object):
     """
     A "worker" process that continually looks for available messages to run
@@ -173,10 +181,11 @@ class WorkerGroup(Worker):
         msg_prefix = "Shutdown signal received by WorkerGroup, "
         if self.stop_gracefully:
             logger.info(msg_prefix + "waiting for sub-workers.")
-            self.wait_for_workers()
+            if not self.in_job:
+                raise StopWorkerGroupLoop()
         else:
             logger.info(msg_prefix + "terminating immediately.")
-        sys.exit(0)
+            sys.exit(0)
 
     def ready(self):
         super(WorkerGroup, self).ready()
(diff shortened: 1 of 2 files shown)


tests/test_worker.py
```diff
@@ -124,8 +129,14 @@ def callback(channel, message):
         worker_group_t.start()
         # wait when a worker starts the callback and terminate the worker group
         callback_is_running.wait()
-        self.assertRaises(SystemExit, worker_group.sigterm_handler, None, None)
-        self.assertFalse(callback_is_running.is_set())
-
+        try:
+            worker_group.sigterm_handler(None, None)
+        except StopWorkerGroupLoop:
+            pass
```
proofit404: If this exception is always propagated, should we use `assertRaises` here?

vshlapakov: Well, the thing is that it's [not always propagated](https://travis-ci.org/django/channels/jobs/308991215). The problem here is that *usually* sub-worker threads start earlier and pick a message, so main worker is idle when we call sigterm handler and almost always there's an exception to break its loop; but [*sometimes*](https://github.com/django/channels/pull/792/files#diff-28a43fbdfeeb97555c5e29e15383a927R184) main worker picks it by itself, and calling sigterm handler doesn't bring the exception - we're humbly waiting when message is processed and the exception is not needed.

 This looks a bit tricky, so I improved the tests to cover all possible cases, please take a look.

channels/worker.py
```diff
@@ -85,11 +93,11 @@ def run(self):
         while not self.termed:
             self.in_job = False
             channel, content = self.channel_layer.receive_many(channels, block=True)
-            self.in_job = True
             # If no message, stall a little to avoid busy-looping then continue
             if channel is None:
                 time.sleep(0.01)
                 continue
+            self.in_job = True
```
vshlapakov: I moved setting the flag here because getting `(channel, content)` doesn't always means that the worker get a message and have something to process, I encountered it when covering a multiple-workers case with tests, let me know if you disagree and I'm missing something here.

Issue #795, opened 2017-11-29
[channels2] Consumer receive function


brianmay:
In attempting to get multiplexed websockets working (I believe I have it "almost working" :-) ), I noticed the following problem:

```
  File "/Users/brianmay/tree/myro/channels/channels/generic/websockets.py", line 40, in websocket_receive
    self.receive(text=message['text'])
  receive() missing 1 required positional argument: 'content'
```

In `WebsocketConsumer` we define receive as:

```
    def receive(self, text=None, bytes=None):                                    
```

(the use of the name bytes is unfortunate :-( - we replace the built-in Python type)

The problem being when override `WebsocketConsumer` with `JsonWebsocketConsumer` we also change the parameters:

```
    def receive(self, content, **kwargs):
```

The function call doesn't provide the required content parameter.

As it is, I think we need two functions, one that takes text/bytes, and the other one takes the JSON content.

Maybe the ``raw_receive()`` function was somehow intended to fix this?

andrewgodwin:
The idea of `raw_receive` was that you could override `receive` and not have to re-implement other stuff, but I think we could do with some reworking around this stuff. I might say call the second one `receive_json`, and probably now is the time to not call it `bytes` too.

brianmay:
Any suggested alternatives to `bytes`? I am hopeless at naming things :-(

andrewgodwin:
Well the underlying websocket concept is called a `binary frame`, but I didn't want to call it `binary` as that is perhaps even more misleading. Maybe `bytestring`? `byte_data` and `text_data`?

brianmay:
`byte_data` and `text_data` seem fine to me.

andrewgodwin:
Maybe even `bytes_data`. That seems to flow better somehow.

brianmay:
Yes, agreed.

PR #796, opened 2017-11-30
Fix issues surrounding multiplexers

brianmay:
Very rough. Not suitable for merging yet. Some known problems.

Perhaps the biggest change is that I try to keep an instance for each consumer class recorded for each consumer instance, so it is possible to store per connection state variables.

In particular, if there are multiple consumers and they are derived from `WebsocketConsumer`, each one will call `accept()` when `connect()` is called. This is invalid, we can only call `accept()` once. 

This change also sets the `base_send` method in the consumer. This was so the call the `accept()` wouldn't crash when it tries to send the acceptance packet. This feels like the wrong solution to me.

I am wondering if maybe we need a new base class for multiplexer consumers, as a lot of the functionality for `WebsocketConsumer` is not really relevant anymore (at least in this implementation). Especially the sending stuff.

In my code, I have overridden the `connect()` method of the consumers not to call its super method. It works, but I don't consider this a good solution.

I also feel uncomfortable removing the error conditions on calling the `send()` and `group_send()` methods in the `Demultiplexer`, however, these methods are currently used for sending responses (assuming I have understood how to do this correctly).

I haven't extensively tested this yet to ensure it does the right thing, but it does seem to run without errors now. :-)

Still needed, but not implemented:

* Ability find Django user associated with websocket.
* Identical stuff with async.

### DIFF
--- a/channels/consumer.py
+++ b/channels/consumer.py
@@ -83,8 +83,7 @@ class SyncConsumer(AsyncConsumer):
 
     _sync = True
 
-    @sync_to_async
-    def dispatch(self, message):
+    async def dispatch(self, message):
         """
         Dispatches incoming messages to type-based handlers asynchronously.
         """
--- a/channels/generic/websockets.py
+++ b/channels/generic/websockets.py
@@ -97,23 +97,26 @@ class JsonWebsocketConsumer(WebsocketConsumer):
     error on binary data.
     """
 
-    def raw_receive(self, message, **kwargs):
-        if "text" in message:
-            self.receive(self.decode_json(message['text']), **kwargs)
+    def receive(self, text=None, bytes=None):
+        """
+        Called with a decoded WebSocket frame.
+        """
+        if text is not None:
+            self.receive_json(self.decode_json(text), **kwargs)
         else:
             raise ValueError("No text section for incoming WebSocket frame!")
 
-    def receive(self, content, **kwargs):
+    def receive_json(self, content):
         """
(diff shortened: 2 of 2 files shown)


andrewgodwin:
Couple of quick comments:

- I agree trying to base the multiplexer on WebsocketConsumer is probably not worth it. It should probably be its own separate ASGI app, which means it will be able to absorb multiple accepts being sent etc.

- Django user/auth stuff is coming soon, that's on my rough plan for the next couple of weeks, along with a Channels test suite that actually works etc.

brianmay:
This fixes #795.

andrewgodwin:
Just FYI, I might have to change all of the multiplexer code significantly soon. I'll try and keep this PR in mind but since it's not mergeable right now I can't guarantee.

brianmay:
No problem. This was more an exercise for me to try and understand what is happening in channels. I can try to keep this up-to-date with any changes you might make - assuming of course - any of this remains relevant.

Not sure what you mean by not mergeable. The tests fail, but I think that is because the tests were broken in the base version.

andrewgodwin:
Yeah, I forgot the tests were still broken on Travis (they work locally these days). I'll try to get that fixed soonish.

andrewgodwin:
OK, all the tests are good these days and the 2.0 release just went out, so things should be in a good place to get this updated and look towards landing it.

brianmay:
Looks like this PR has been superseded by #825 :-). Will close.

PR #797, opened 2017-11-30
Rename text/bytes parameters

brianmay:
This change is to ensure the bytes parameter doesn't conflict with the
Python built in type.


brianmay:
As discussed in #795. Note that I did not change the messages.

PR #803, opened 2017-12-07
Make Response skippable, also with middleware

ivorbosloper:


andrewgodwin:
Can you pick a response code in the `1xx` range that isn't 100? I'd rather not step on an existing HTTP code. Maybe `113` or something random?

ivorbosloper:
@andrewgodwin I've changed it to 113.

andrewgodwin:
Thanks! I'll get round to putting it in a release soon.

PR #805, opened 2017-12-09
Use serializer better on bindings (a.k.a. rename "data" to "fields" on data binding)

valberg:
I wanted to provide initial data from my web app using an "initial" stream like so:

```
class TodoBinding(WebsocketBinding):
    model = Todo
    stream = "todo"
    fields = ["id", "description", "is_done"]

    @classmethod
    def group_names(cls, instance):
        return ["todo"]

    def has_permission(self, user, action, pk):
        return True


class InitialDataConsumer(JsonWebsocketConsumer):
    def connect(self, message, multiplexer, **kwargs):
        todos = Todo.objects.all()
        serialized_todos = serializers.serialize(
            'python', 
            todos
        )
        multiplexer.send(serialized_todos)


class Demultiplexer(WebsocketDemultiplexer):
    consumers = {
        "initial": InitialDataConsumer,
        "todo": TodoBinding.consumer
    }

    def connection_groups(self):
        return ["todo"]

```

This way I would get data like this on connection:

```
{
    "stream": "initial", 
    "payload": [
        {
            "model": "todo.todo", 
            "pk": "<UUID-string>", 
            "fields": {
                "description": "Write some awesome apps using django channels.",
                "is_done": false
            }
        },
        ...
    ]
}
```

and like this on, for instance, creation of a todo:

```
{
	"stream": "todo", 
	"payload": {
		"model": "todo.todo", 
		"pk": "<UUID-string>", 
		"data": {
			"description": "New todo", 
			"is_done": false
		}, 
		"action": "create"
	}
}
```

The problem here is that the shape of the data is almost the same, only the name of "fields" and "data" differ.

So the patch is about making channels use serializers without renaming the fields key-value-pair and as a result hereof rename "data" to "fields" in both the outbound and inbound data format.

I might be overseeing something, so just shoot with questions and such :)

### DIFF
--- a/channels/binding/websockets.py
+++ b/channels/binding/websockets.py
@@ -92,7 +92,7 @@ class WebsocketBinding(Binding):
         body = json.loads(message['text'])
         action = body['action']
         pk = body.get('pk', None)
-        data = body.get('data', None)
+        data = body.get('fields', None)
         return action, pk, data
 
     def _hydrate(self, pk, data):
--- a/channels/binding/websockets.py
+++ b/channels/binding/websockets.py
@@ -41,12 +41,8 @@ class WebsocketBinding(Binding):
         return WebsocketMultiplexer.encode(stream, payload)
 
     def serialize(self, instance, action):
-        payload = {
-            "action": action,
-            "pk": instance.pk,
-            "data": self.serialize_data(instance),
-            "model": self.model_label,
-        }
+        payload = self.serialize_data(instance)
+        payload["action"] = action
         return payload
 
     def serialize_data(self, instance):
@@ -60,8 +56,7 @@ class WebsocketBinding(Binding):
                 fields = self.fields
         else:
             fields = [f.name for f in instance._meta.get_fields() if f.name not in self.exclude]
-        data = serializers.serialize('json', [instance], fields=fields)
-        return json.loads(data)[0]['fields']
+        return serializers.serialize('python', [instance], fields=fields)[0]
 
     # Inbound
     @classmethod
--- a/tests/test_binding.py
+++ b/tests/test_binding.py
@@ -36,11 +36,11 @@ class TestsBinding(ChannelTestCase):
         received = client.receive()
         self.assertTrue('payload' in received)
         self.assertTrue('action' in received['payload'])
-        self.assertTrue('data' in received['payload'])
-        self.assertTrue('username' in received['payload']['data'])
-        self.assertTrue('email' in received['payload']['data'])
-        self.assertTrue('password' in received['payload']['data'])
-        self.assertTrue('last_name' in received['payload']['data'])
+        self.assertTrue('fields' in received['payload'])
+        self.assertTrue('username' in received['payload']['fields'])
+        self.assertTrue('email' in received['payload']['fields'])
+        self.assertTrue('password' in received['payload']['fields'])
+        self.assertTrue('last_name' in received['payload']['fields'])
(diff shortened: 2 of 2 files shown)


proofit404:
Hi, thanks for your contribution.

Correct me if I'm wrong but this PR change is a matter of taste and doesn't include any new functionality. Also, this change doesn't look backward compatible to me.

Why you can't use old-one format? `fields` were there for model binding meaning like "notify me about model change on this fields".

Closing for now. Feel free to reopen if you think it worth further discussion.

Regards, Artem.

valberg:
Hey @proofit404,

It is all about what `django.core.serializers.serialize()` returns. Currently `channels.binding.websockets.WebsocketBinding.serialize()` is doing double work for something that is already done by `django.core.serializers.serialize()`, the only difference is that we call the "fields" key for "data". We're even adding `pk` and `model` which were already provided by `django.core.serializers.serialize()`.

My use case is that I'm writing a library for Elm for dealing with communicating with django channels via websockets. Elm is typed, so if I want to have frontend code that deals with both bindings and the initial data, the data sent has to have the same shape (including same keys in the JSON object) in both cases. For that to happen I can do one of three things:

0. Implement my `InitialDataConsumer` as follows (as you can see the iteration of `serialized_todos` just to rename `fields` to `data` seems a bit silly :)

```
class InitialDataConsumer(JsonWebsocketConsumer):
    def connect(self, message, multiplexer, **kwargs):
        todos = Todo.objects.all()
        serialized_todos = serializers.serialize(
            'python', 
            todos
        )
        # Change 'fields' to 'data' in the resulting output
        for todo in serialized_todos:
            todo['data'] = todo.pop('fields')
        multiplexer.send(serialized_todos)
```

1. Rely on that channels uses basic `django.core.serializers` functionality without any changes to the output (what this PR achieves).

2. Extend `channels.binding.websockets.WebsocketBinding` so it is possible to extend the consumer generated with data that should be sent upon connecting, this way it is possible to reuse `channels.binding.websockets.WebsocketBinding.serialize_data()` when serializing the initial data.

There is no surprise that I'm in favor of number 1 (and number 2 is also quite interesting, but I thought I'd try the path of "just use the output of `django.core.serializers.serialize()`" first :)

I'm up for discussing this further, since I think there is some value in having a sane API for dealing with initial data and subsequent "data binding actions" in the same way.

Does what I'm getting at make sense or does it still smell like "matter of taste"? :)

andrewgodwin:
Hey @valberg - while I think matter of taste is the wrong word, the binding framework provided in the base Channels package is deliberately very simplistic especially in regards to serialisation, so that's one reason we may not consider changes. I would love a better solution that included initial data and updates, and maybe even proper state tracking on a backend, but that belongs in a third-party package (I may even move the binding we have now into a third-party package for channels 2 because I'm not sure it belongs in the core)

The main problem here, however, is that this isn't backwards-compatible, which would be a hard requirement. Channels 2 would potentially provide a way to jump this as it has breaking changes, but I am also keen to keep those changes minimal. Instead, I recommend building on top of the existing/new classes (depending on which version you want to target and when I manage to get it into beta) and maintain things as a third-party app that extends or replaces core classes.

valberg:
Hey @andrewgodwin,

That makes sense. I'll look into writing a third-party app with bindings as you suggest.

Thanks both @proofit404 and @andrewgodwin for taking your time to read through my thoughts and ideas :) Your work is highly appreciated!

Issue #195, opened 2016-06-03
enforce_ordering message dropping should close a WebSocket


trainman419:
If I use `enforce_ordering` and I send messages over a websocket more quickly than my consumer can process them, eventually the `__wait__` channel becomes full and starts dropping messages.

Since `enforce_ordering` expects messages with sequential IDs, if a message gets dropped it causes the enforce_ordering filter to wait forever for that message.

The stack trace that I get when this happens is:

```
2016-06-03 18:25:37,917 - ERROR - worker - Error processing message with consumer model.consumers.ws_receive:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/channels/worker.py", line 106, in run
    consumer(message, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/channels/sessions.py", line 64, in inner
    return func(message, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/channels/sessions.py", line 127, in inner
    "full __wait__ channel for socket %s" % message.reply_channel.name
ChannelFull: Cannot add unordered message to already full __wait__ channel for socket websocket.send!XzbYSOku
```

andrewgodwin:
Hmm, this is going to happen - with enforce_ordering on we have to drop whole sockets if you get lagged behind - but it should just do that (disconnect the socket) rather than tracebacking.

trainman419:
I'm willing to try to put together a simple test case that reproduces this, but I'm new to django and channels and I'm not sure which tools to use. Is there any documentation on the subject that you can point me to?

andrewgodwin:
There's not a lot about writing tests against Channels itself; you'd have to take a look at the existing test suite if you're interested, and I'm not sure that has the example of setting a low channel capacity and testing channel decorators combined.

andrewgodwin:
Renamed this to better describe what it is, which is that we need to have a way to close WebSockets when their messages get dropped due to ordering violations, but not have the channels code specifically looking out for websockets.

My current suggestion would be that if a message is dropped by ordering (or something else) and channels knows it's definitely gone, a standardised "control" message reply is sent back to the `reply_channel` if one is present saying that it was dropped.

We'd need to make the worker do this, specify what the control message looks like in the asgi spec, and then have Daphne respect that message and close websockets.

asgorh:
Is there any chance that this issue will be fixed in near future?

Regards
Tom

andrewgodwin:
[...]

PR #810, opened 2017-12-22
Updated enforce_ordering example to include decorator on connect

sachinrekhi:
Updated the documentation for enforce ordering to add the enforce_ordering decorator to the connect consumer.

Discussion here: https://github.com/django/channels/issues/195


proofit404:
Thanks!

PR #823, opened 2018-01-18
Fix receive issues with WebsocketConsumer and JSONWebsocketConsumer.

### LINKED ISSUE #795: [channels2] Consumer receive function

bpeschier:
This fixes #795 and contains a part of #796; it only deals with WebsocketConsumer and JSONWebsocketConsumer's `(raw_)receive(_json)`. It also renames `send()` to `send_json()` and updates the docs accordingly.

I know this might all change in the near future, but it makes it possible to toy around with 2.0 at the moment! :)

### DIFF
--- a/channels/generic/websockets.py
+++ b/channels/generic/websockets.py
@@ -51,7 +51,7 @@ class WebsocketConsumer(SyncConsumer):
         """
         Sends a reply back down the WebSocket
         """
-        if text is not None:
+        if text_data is not None:
             super(WebsocketConsumer, self).send(
                 {"type": "websocket.send", "text": text_data},
             )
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Yes, this is a much nicer way of handling it. Good to get this in before release.

PR #827, opened 2018-01-21
Fixed missing argument in channel layer docs.

apollo13:


apollo13:
Sorry, I did this patch on a bus and didn't realize that it created the commit in the channels repo itself; will do better next time :)

andrewgodwin:
That's fine, you have commit after all :)

PR #831, opened 2018-01-25
[JS client] fix callback parameter name : action => payload

pchiquet:
this variable is set to "msg.payload" when a message is received


proofit404:
Hi

I wonder if it is just a variable name change?

andrewgodwin:
This patch doesn't seem to have any actual external effect - did you just want to rename it internally?

Also, if you're submitting a JavaScript PR, please recompile the built versions and include those in the diff as we don't have a build process for JS.

pchiquet:
Hi,
- it is only a variable name change in the callback called when any websocket message is received (this callback passes the full message payload, and not "payload.action")
- I have updated the README to reflect these changes
- I have recompiled with "npm run compile"
- the tests pass (launched with "npm run test")

andrewgodwin:
Thanks!

PR #847, opened 2018-02-03
Fixing small irragularities in example consumer

valberg:
This confused me a bit to begin with until I checked the actual source of `WebSocketConsumer`.


andrewgodwin:
And here you can see the things that changed during development! Thanks :)

valberg:
You're welcome! Awesome work btw! Looking forward to giving it a spin in the wild! 😄

PR #862, opened 2018-02-05
doc: authentication change username to user.

Christophe31:
As it seems to be a user object, reading the doc made me think it was only a string (username)


andrewgodwin:
Yup, good catch! Thanks.

PR #863, opened 2018-02-05
Full in memory channel layer 

svenauhagen:
Hi,

I added the full in memory layer back to channels 2.
We always used it in channels 1 for local testing.

Can you merge it into the code?

Best and thanks
Sven

### DIFF
--- a/channels/layers.py
+++ b/channels/layers.py
@@ -5,10 +5,17 @@ import fnmatch
 import random
 import re
 import string
+import time
+import threading
 
 from django.conf import settings
 from django.utils.module_loading import import_string
+from django.utils.translation import ugettext as _
 
+from collections import deque
+from copy import deepcopy
+
+from channels.exceptions import ChannelFull, InvalidChannelLayerError
 from channels import DEFAULT_CHANNEL_LAYER
 
 from .exceptions import InvalidChannelLayerError
@@ -177,9 +184,21 @@ class InMemoryChannelLayer(BaseChannelLayer):
     In-memory channel layer implementation for testing purposes.
     """
 
-    def __init__(self, *args, **kwargs):
-        super(InMemoryChannelLayer, self).__init__(*args, **kwargs)
+    local_poll_interval = 0.01
+
+    def __init__(self, expiry=60, capacity=100, channel_capacity=None, **kwargs):
+        super().__init__(expiry=expiry, capacity=capacity, channel_capacity=channel_capacity, **kwargs)
         self.channels = {}
+        self.groups = {}
+        self.thread_lock = threading.Lock()
+        # Coroutines currently receiving the process-local channel.
+        self.receive_tasks = {}
+        # Buffered messages by process-local channel name
+        self.receive_buffer = {}
+
+    ### Channel layer API ###
+
+    extensions = ["groups", "flush"]
 
     async def send(self, channel, message):
         """
@@ -194,8 +213,20 @@ class InMemoryChannelLayer(BaseChannelLayer):
             message = dict(message.items())
             message["__asgi_channel__"] = channel
             channel = self.non_local_name(channel)
-        # Store it in our channels list
-        self.channels.setdefault(channel, []).append(message)
+            # Store it in our channels list
+            queue = self.channels.setdefault(channel, deque())
+            # Are we full
+            if len(queue) >= self.capacity:
+                raise ChannelFull(channel)
+
+            # Add message
+            queue.append((
+                time.time() + self.expiry,
+                deepcopy(message),
+            ))
+        else:
+            # We do not support non local channels
+            pass
 
     async def receive(self, channel):
         """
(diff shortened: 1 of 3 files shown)


andrewgodwin:
I will happily take this once it has unit tests (I'll write the documentation part) - you may want to reuse the `channels_redis` unit tests for the most part.

svenauhagen:
I ported over the channel_redis tests to the inmemory layer.
Let me know if that is alright or if you want me to add additional tests.

svenauhagen:
Ah yes, I wasn't aware of the asyncio queue.
I am knew to asyncio just learned it over the weekend to write the memory channel layer :)

Do I understand you correctly that I should simply use the raw channel name for the queue key?
Should I also dismiss the assert self.valid_channel_name(channel, receive=True) check then?

andrewgodwin:
You can leave the `valid` check in, but there's no need to do the part where it takes a name with a `!` in it and splits it into a non-local part, and puts the real channel name into the message dict.

svenauhagen:
Alright, can you look at it now.
I tested it today and it worked well that way.

channels/layers.py
```diff
@@ -173,13 +176,20 @@ def non_local_name(self, name):
 
 
 class InMemoryChannelLayer(BaseChannelLayer):
-    """
-    In-memory channel layer implementation for testing purposes.
-    """
+    '''
+    Our own in memory layer
```
brianmay: Did you mean to change the docstring?

svenauhagen: No that was left in there from testing I changed it to "In-memory channel layer implementation" since it can be used for more than testing now

channels/layers.py
```diff
@@ -204,13 +219,14 @@ def __init__(self, *args, **kwargs):
...
-            except IndexError:
-                asyncio.sleep(0.01)
-            else:
-                return message
+        self._clean_expired()
+
+        queue = self.channels.setdefault(channel, asyncio.Queue())
+
+        # Do a plain direct receive
+        _, message = await queue.get()
```
andrewgodwin: I would add something here to clean up the queue entry in the `self.channels` dict if it's now empty.

channels/layers.py
```diff
@@ -173,13 +176,20 @@ def non_local_name(self, name):
 
 
 class InMemoryChannelLayer(BaseChannelLayer):
-    """
-    In-memory channel layer implementation for testing purposes.
-    """
+    '''
+    In-memory channel layer implementation
+    '''
```
andrewgodwin: Please change these back to `"""` - trying to keep everything consistent!

channels/layers.py
```diff
@@ -173,13 +176,20 @@ def non_local_name(self, name):
...
+    local_poll_interval = 0.01
 
-    def __init__(self, *args, **kwargs):
-        super(InMemoryChannelLayer, self).__init__(*args, **kwargs)
+    def __init__(self, expiry=60, capacity=100, channel_capacity=None, **kwargs):
+        super().__init__(expiry=expiry, capacity=capacity, channel_capacity=channel_capacity, **kwargs)
         self.channels = {}
+        self.groups = {}
+        self.thread_lock = threading.Lock()
```
andrewgodwin: I'm not sure a `threading.Lock` makes sense here - you shouldn't need it since you're writing async code. Did you find a reason to add it?

channels/layers.py
```diff
@@ -222,6 +238,85 @@ def __init__(self, *args, **kwargs):
...
+                # See if it's expired
+                while not queue.empty() and queue._queue[0][0] < time.time():
+                    queue.get_nowait()
+                    remove = True
+                # Any removal prompts group discard
+                if remove:
+                    self._remove_from_groups(channel)
+                # Is the channel now empty and needs deleting?
+                if not queue:
+                    del self.channels[channel]
```
andrewgodwin: This method should also clean expired group memberships separately.

svenauhagen:
Great, I will get on these cleanups later today.
Regarding the thread lock I do have a general question since I am not sure.
How is daphne handling new web requests in 2.0?
Is it spawning a thread for each request or is it entirely done with asyncio coroutines?

andrewgodwin:
It's all coroutines. Some requests will end up using threads if they're based on `SyncConsumer`, but all the async stuff (calls to channel layers included) happens in a single main thread on a single event loop.

svenauhagen:
Thanks for the explanation.
Isn't that a problem though if some async requests run longer without await daphne will be blocked to answer new requests?

andrewgodwin:
No, because they're async requests, as long as they `await` anything that's long running (as everything is, like receiving off the channel layer) the event loop frees up. You can have thousands of coroutines in the same event loop easily.

channels/layers.py
svenauhagen: What do you mean by that exactly?
Should I remove the channel from all groups if not queue?

andrewgodwin: You need to go through the group memberships and expire the entries if their expiry time is bigger than `group_expiry` (there should be a test for this too, but I'll add that later). The Redis one does it here: https://github.com/django/channels_redis/blob/master/channels_redis/core.py#L316

svenauhagen:
Ok I think I got it now

channels/layers.py
```diff
@@ -267,7 +266,8 @@ def _clean_expired(self):
         for group in self.groups:
             for channel in self.groups.get(group, set()):
                 # If join time is older than group_expiry end the group membership
-                if self.groups[group][channel] and int(self.groups[group][channel]) < (int(time.time()) - self.group_expiry):
+                if (self.groups[group][channel] and
+                    int(self.groups[group][channel]) < (int(time.time()) - self.group_expiry)):
                     del self.groups[group][channel]
```
proofit404: Maybe shortcut local variable will be better.

svenauhagen: I added a timeout variable

andrewgodwin:
Looks good! I am ready to merge this if you are,

brubbel:
@svenauhagen @andrewgodwin 
https://github.com/django/channels/blob/d6643d96dd59e153cda13681a6d5a8963b9d577c/channels/layers.py#L268-L272

In python3, looping over a dict/set defaults to iterators, and _clean_expired() raises an exception, as changing the dict/set is not allowed during iteration.
`for channel in self.groups.get(group, set()):`
should be
`for channel in list(self.groups.get(group, set())):`

Error below:
```
Exception in thread Thread-5:
  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "/home/xyz/venv3/lib/python3.5/site-packages/asgiref/sync.py", line 49, in __call__
    return call_result.result()
  File "/usr/lib/python3.5/concurrent/futures/_base.py", line 405, in result
    return self.__get_result()
  File "/usr/lib/python3.5/concurrent/futures/_base.py", line 357, in __get_result
    raise self._exception
  File "/home/xyz/venv3/lib/python3.5/site-packages/asgiref/sync.py", line 63, in main_wrap
  File "/home/xyz/django/abc/consumers.py", line 47, in sender
    await self.channel_layer.group_send(self.MY_GROUP, message)
  File "/home/xyz/venv3/lib/python3.5/site-packages/channels/channels/layers.py", line 321, in group_send
    self._clean_expired()
  File "/home/xyz/venv3/lib/python3.5/site-packages/channels/channels/layers.py", line 268, in _clean_expired
    for channel in self.groups.get(group, set()):
RuntimeError: dictionary changed size during iteration
```

andrewgodwin:
I just pushed up a commit to fix that.

brubbel:
I want to point out another (since longtime) issue with the _clean_expired() method. I've noticed that channels won't always get cleaned up, and _clean_expired() just keeps 'wasting' cpu cycles.

What I _think_ is the reason:

For example, a channel becomes 'orphan' due to a broken websocket connection, and there are _regular transmissions_ to this orphan channel (e.g. group_send() broadcasts).

The _clean_expired() for this channel is now only called on a group_send(), which adds a fresh message to the queue. The following loop removes the expired messages:
```
            while not queue.empty() and queue._queue[0][0] < time.time():
                queue.get_nowait()
                remove = True
```

However, the channel is never deleted, because a channel is only cleaned up when the queue is empty: ```
while group_send() keeps filling the queue:
            # Is the channel now empty and needs deleting?
            if not queue:
                del self.channels[channel]
```
And regular broadcasts prevent the queue to become empty.

Affected code:
https://github.com/django/channels/blob/b58c213aaf507e555c88a667f722e3e0a24f4ff9/channels/layers.py#L252-L263

**Comment by brubbel** [comment]
I would like to point out some additional issues.
https://github.com/django/channels/blob/b58c213aaf507e555c88a667f722e3e0a24f4ff9/channels/layers.py#L261-L263
Previously (before asyncio), self.channels had been using a collections.deque, for which the following is valid to check if queue is empty:
```
if not queue:
     do_something()
```
However, the truth value of an queue.Queue or asyncio.Queue cannot be used to check whether the queue is empty. It will always return _True_ and the channel is never deleted. The correct statement is as follows:
```
if queue.empty():
     do_something()
```

brubbel:
Even with the above fixes, the InMemoryChannelLayer keeps leaking channel objects in self.channels(), when websocket connections are coming and going over time. The group_discard() method is called, which is correct, but channels with an empty queue keep hanging around.

I suggest the following fixes to _group_expire(), please comment if I overlooked something:
https://github.com/brubbel/channels/commit/ad6b135f9dd29036fbedc0004003941edaf480c2

andrewgodwin:
Can you open a separate issue to track this? It's getting complex enough it needs one.

It's also worth noting that the in-memory channel layer is really only meant for testing, so fixes to it are going to be lower priority than some other bugs in the queue!

brubbel:
I am fully aware that the in-memory channel layer is for debug only, however -imho- for applications on a lightweight system such as raspberry pi, I think it is a reasonable alternative.

andrewgodwin:
Right, just saying that fixing this is going to come below the 6 or so other bugs I have to look at at the moment!

PR #865, opened 2018-02-05
Fix server error: 'dict' object has no attribute 'content'

shtalinberg:
In Chrome with AllowedHostsOnlyOriginValidator I get 
Request Method:GET
Status Code:500 Internal server error: 'dict' object has no attribute 'content'
with my changes all work fine

### DIFF
--- a/channels/security/websockets.py
+++ b/channels/security/websockets.py
@@ -35,7 +35,11 @@ class BaseOriginValidator(object):
         return self.func(message, *args, **kwargs)
 
     def get_header(self, message, name):
-        headers = message.content["headers"]
+        if hasattr(message, "content"):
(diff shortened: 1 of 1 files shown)


andrewgodwin:
The file you are trying to patch is gone in the master branch and replaced by https://github.com/django/channels/blob/master/channels/security/websocket.py which should not have this problem.

shtalinberg:
But in 2.0.0 release - file with this problem. 
I was think documentation is wrong -  it's code in 2.0.0 wrong ?

andrewgodwin:
Yes, the 2.0.0 release has this file, but it's been fixed on the `master` branch already, and I'll be doing a new `2.0.1` release with the fix in it tonight.

PR #867, opened 2018-02-05
Fix URLRoute resolve match

svenauhagen:
Hi,

this is a fix for the URLRouter.
When executing the for loop the line match = route.resolve(path) throws an exception if the path does not match.
This will break the for loop and therefore other possible matches are not checked.
The change will catch that exception to continue the loop.

Best
Sven


andrewgodwin:
Thanks, not sure how I missed it. I won't make you write a test as I should have done that.

PR #869, opened 2018-02-05
Add missing import in authentication example

orn688:
`url` was not imported from `django.conf.urls` in the first authentication example in the docs, which seems like it was a mistake – the code doesn't work as-is.


andrewgodwin:
Thanks!

PR #873, opened 2018-02-06
await disconnect and wait

dgilge:
When I understand it correctly you should `await` `communicator.disconnect()` and `communicator.wait()` as you do it in the [tests](https://github.com/django/channels/blob/master/tests/test_generic_websocket.py).


proofit404:
Thanks!

PR #874, opened 2018-02-06
Update example ChatConsumer method signature

acdha:
This avoids the unexpected keyword argument for `text_data`. Since Channels only supports Python 3.5 now, I also made it a keyword-only argument.


andrewgodwin:
Thanks!

PR #875, opened 2018-02-06
Update PingConsumer to minimal working class

acdha:
This adds extra code but I think it's useful for the examples to work out-of-the box.


proofit404:
Thanks!

andrewgodwin:
Ah yes, thanks. We can probably change that example to use `AsyncWebsocketConsumer` now it's written as that'll accept by default.

PR #884, opened 2018-02-08
Add Asyncio Childwatcher to main thread

svenauhagen:
Hi,

I want to use asyncio.create_subprocess_exec which allows me to start processes and pass along its  output to a channel.
It does need a childwatcher which must be initialized in the main thread.
I added the code in the runserver.py for running it locally, can you merge it?

Thanks and best
Sven

### DIFF
--- a/channels/management/commands/runserver.py
+++ b/channels/management/commands/runserver.py
@@ -1,3 +1,4 @@
+import asyncio
 import datetime
 import sys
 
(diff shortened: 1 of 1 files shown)


andrewgodwin:
This is not something I can merge into Channels, as you'd also need it in Daphne. I instead suggest that you add code to one of your apps' setup() methods, or have code from where you launch a child process that finds the main thread and calls this.

PR #886, opened 2018-02-08
Permissions Mixins

svenauhagen:
Hi,

these are two permission mixins for the websocketconsumer classes to check user permissions on connect.

I am not sure how to test that, more precise how I can create a testserver to start a consumer to check for this in a test.

Best
Sven

### DIFF
--- a/channels/generic/mixins.py
+++ b/channels/generic/mixins.py
@@ -2,7 +2,7 @@ class PermissionsMixin(object):
     """
     Mixin to check if user has the correct permissions
     """
-    perm = 'consumer.changeme'
+    perm = "consumer.changeme"
 
     def websocket_connect(self, message):
         if self.has_permission():
@@ -25,7 +25,7 @@ class AsyncPermissionsMixin(object):
     """
     Mixin to check if user has the correct permissions
     """
-    perm = 'consumer.changeme'
+    perm = "consumer.changeme"
 
     async def websocket_connect(self, message):
         if self.has_permission():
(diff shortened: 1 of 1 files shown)


channels/generic/mixins.py
```diff
@@ -0,0 +1,44 @@
+class PermissionsMixin(object):
```
auvipy: explicit inherit from object is not needed in python3

channels/generic/mixins.py
```diff
@@ -0,0 +1,44 @@
...
+        Check for connect permissions
+        '''
+        if self.perm:
+            if self.scope["user"] and self.scope["user"].is_authenticated and self.scope["user"].has_perms(self.perm):
+                return True
+
+        return False
+
+
+class AsyncPermissionsMixin(object):
```
auvipy: same goes here..

svenauhagen:
Thanks got it

andrewgodwin:
Thanks for the idea, but I'm afraid I'm not going to accept these in - first of all, I don't think mixins make for good user-facing APIs, and I'd rather these were written as a simple `check_permissions` function that raised `AcceptConnction` or `DenyConnection`.

Secondly, this is the sort of code I don't think belongs in Channels itself as it's too specific to one way of implementation, but if you like, you should start a third-party app with pluggable features like this (as you would do for Django itself).

svenauhagen:
Got it thanks. I am missing a permission check for consumers to get a fine grain check per user for my projects. Would be great to see something like that in channels in the future.

andrewgodwin:
We already give you a `connect()` method where you can easily do a permission check or whatever check you want to let a user in; I doubt we will go beyond that in the near future. Keeping the core code simple and maintainable requires not extending too far, sorry!

PR #887, opened 2018-02-08
Various docs fixes

laevilgenius:
Also [one-to-two.rst#group-objects](https://github.com/django/channels/blob/fc2fb74740c3d652ef89d06b691d011857e6a68e/docs/one-to-two.rst#group-objects) looks weird: as if AsyncConsumer and SyncConsumer were mixed.


andrewgodwin:
Thanks! And yes, that other section is weird, I'll update it now.

PR #888, opened 2018-02-08
Add StaticFilesWrapper to ChannelsLiveServerTestCase

### LINKED ISSUE #880: ChannelsLiveServerTestCase doesn't have staticfiles support
It needs a class attribute that lets you turn it on, and sets the ASGI application to `StaticFilesWrapper(get_default_application())`

dgilge:
This fixes #880

**Some notes:**

Maybe it is better to serve static files by default because you will need it in most cases in order to serve some JavaScript library.

I didn't check `apps.is_installed("django.contrib.staticfiles")` because I think you will use static files in your tests only if you installed `staticfiles`. Or shall I raise an `ImproperlyConfigured` if you try to serve static files in the tests without installed `staticfiles` before?

I tried to copy the tests in question from Django but I saw that there are multiple files involved. Therefore I stopped my efforts to include them in this PR.

### DIFF
--- a/channels/testing/live.py
+++ b/channels/testing/live.py
@@ -72,7 +72,7 @@ class ChannelsLiveServerTestCase(LiveServerTestCase):
 
     In order to serve static files create a subclass with serve_static = True.
     """
-    
+
     server_thread_class = DaphneServerThread
     static_wrapper = StaticFilesWrapper
     serve_static = False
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I'm fine with the lack of detection of the app being installed. If you fix the linting errors that's making Travis sad then I can merge it in!

dgilge:
Travis is happy now :)

andrewgodwin:
Great - thanks!

PR #896, opened 2018-02-11
Update channel_layers.rst

opmftw:
Just added "close_code" to the disconnect methods, as per the documentation, otherwise an exception is raised - disconnect() takes 1 positional argument but 2 were given, it does not break the code but might be confusing


proofit404:
Thanks!

Issue #825, opened 2018-01-20
Channels 2 Multiplexing


andrewgodwin:
Channels 2 needs multiplexing support, as it's not yet implemented.

brianmay:
I am trying to visualize how this should be implemented.

The problem I keep encountering, however, is that the Demultiplexer needs to be the class that terminates the WebSocket connection. As we don't know where to dispatch messages to until we receive them, and every message for a given connection could have a different destination.

So having the Demultiplexer dispatch to a set of Consumer classes seems wrong, because the Consumer classes each assume that they exclusively terminate the websocket connection. They also contain functionality that is not needed, such as reading/writing from the websocket.

So maybe we need a completely independent consumer class (probably sync and async) dedicated to just receiving Demultiplexer JSON messages?

andrewgodwin:
Yes, this is going to need an ASGI application (probably a consumer but might have to be lower-level) that does the de/re-multiplexing itself and then manages a suite of other application instances, one for each stream. It's going to be a lot more complex than the middlewares/routing that just looks at scope.

hishnash:
You think it needs to be this complex?

Why not do as @brianmay suggested and terminate the WebSocket on the `Demultiplexer` and make the child `Consumer`ish classes not be true `Consumer`s but rather be some `DemultiplexedConsumer` the `Demultiplexer` could keep references to them. And they (the `DemultiplexedConsumer` instances) could have weak refs back to the `Demultiplexer` so as to access session info etc. 

with respect to processing messages coming over layers, the `Demultiplexer` could use the same keys as it does for routing WS messages to rout the layer messages. 

```
class APIDemultiplexer(WebsocketDemultiplexer):
    influencer = InfluencerDemultiplexedConsumer
    user = UserDemultiplexedConsumer
```

sending a message `user.action.logout` would call the method `action_logout` on 
 the instance of `UserDemultiplexedConsumer` for this `Demultiplexer`.

 What would the benefit of `de/re-multiplexing itself and then manages a suite of other application instances` bring?

andrewgodwin:
The whole principle of the new ASGI design is turtles all the way down - that is, a demultiplexer should take any valid ASGI application as its demultiplexing endpoints. Otherwise we would need to maintain a whole separate, parallel suite of generic consumers that are these not-quite-consumers, and you wouldn't be able to move things between them, etc.

The pattern is basically:
* When the multiplexer is initialised it also initialises any sub-applications with the same scope
[...]

PR #900, opened 2018-02-12
[WIP] Re implement de-multiplexing #825

hishnash:
This brings a version of de-multiplexing (for json websockets only) to v2. https://github.com/django/channels/issues/825

this adds 3 classes:

1) `class ApplicationWithChildren:`

(maybe we should find a better name)

this class managed the child Application coroutiens along side the Task/Dispatch for both the upstream and downstream queues for messages. 

2) `class Demultiplexer(ApplicationWithChildren):`

the base class for a `Demultiplexer`, compliant with ASGI application spec.

3) `class AsyncJsonWebsocketDemultiplexer(Demultiplexer, AsyncJsonWebsocketConsumer):`

this works as it used to (more or less, it now supports any valid JSON values for the `payload` since messages are in v2 passed within the same python thread so fewer limitations than before).

and example usage
```python
class MyDemultiplexer(AsyncJsonWebsocketDemultiplexer):
    applications = {
        "user": UserConsumer, 
        "posts": URLRouter([
            url("^/admin/$", AdminPostsConsumer),
            url("^/public/$", PublicPostsConsumer),
        ])
    }
```
---

There is an issue with the cleanup of tasks, https://github.com/hishnash/channels/blob/Re-Implement-Multiplexing-%23825/channels/utils.py#L54-L61  tasks are being cancelled even though the run loop is already closed. 

---

I still need to add Documentation before this is merged just wanted to check that this is a good approach to the problem before it put the time in for those bits.

### DIFF
--- a/channels/generic/multiplexer.py
+++ b/channels/generic/multiplexer.py
@@ -1,0 +1,322 @@
+import asyncio
+import logging
+from functools import partial
+from typing import Any, Dict
+
+from channels.consumer import get_handler_name
+from channels.exceptions import StopConsumer
+from channels.generic.websocket import AsyncJsonWebsocketConsumer
+
+logger = logging.getLogger(__name__)
+
+
+class AsyncJsonWebsocketDemultiplexer(AsyncJsonWebsocketConsumer):
+
+    """
+    JSON-understanding WebSocket consumer subclass that handles demultiplexing
+    streams using a "stream" key in a top-level dict and the actual payload
+    in a sub-dict called "payload". This lets you run multiple streams over
+    a single WebSocket connection in a standardised way.
+    Incoming messages on streams are dispatched to consumers so you can
+    just tie in consumers the normal way.
+
+    Set a mapping of streams to an application classes in the "applications"
+    keyword.
+    """
+
+    # timeout in seconds after close used before killing child applications.
+    application_close_timeout = 10
+
+    # the mapping of applications by stream_name
+    applications = {}  # type: Dict[str: 'typing.Type'[channels.consumer.AsyncConsumer]]
+
+    def __init__(self, scope: Dict[str, Any]):
+        super().__init__(scope)
+        self.stream_upstream_queues = {}  # type: Dict[str, asyncio.Queue]
+        self.accepted_upstream_stream = set()
+        self.disconnecting = False
+        self.sent_close = False
+        self._task_monitors = None  # type: typing.Optional[asyncio.Task]
+
+    # Methods to INTERCEPT Client -> Stream Applications
+
+    async def websocket_connect(self, message):
+        """
+        Connect and then inform each stream application.
+        """
+        await super().websocket_connect(message)
+        for queue in self.stream_upstream_queues.values():
+            queue.put_nowait(message)
+
+    async def websocket_disconnect(self, message):
+        """
+        Handle when the connection is lost.
+        """
+
+        # set the disconnecting so that we know to to send frames further down.
+        self.disconnecting = True
+
+        for stream_name, queue in self.stream_upstream_queues.items():
+            # inform all upstream applications that the connection is going
+            #  down (this includes ones that never accepted!)
+            queue.put_nowait(message)
+
+        if self._task_monitors is None:
+            await super().websocket_disconnect(message)
+            return
+
+        try:
+            # wait for all the upstream applications to close down
+            await asyncio.wait_for(
+                self._task_monitors,
+                timeout=self.application_close_timeout
+            )
+        except asyncio.TimeoutError:
+            # they did not close down fast enough!
+            # not to worry they will be killed very soon.
+            logger.warning(
+                "One or more child application stream of %r took too long to "
+                "shut down and was killed.",
+                self,
+            )
+        # the `disconnect` method on the de-multiplexer is called after
+        # all upstream applications have stopped or the timeout has passed
+        #  (whatever is faster).
+        await super().websocket_disconnect(message)
+
+    async def receive_json(self, content, **kwargs):
+        """
+        Rout the message down the correct stream.
+        """
+        # Check the frame looks good
+        if isinstance(content, dict) and "stream" in content and "payload" in content:
+            # Match it to a channel
+            stream_name = content["stream"]
+            try:
+                stream_queue = await self._get_stream_queue(stream_name)
+
+            except ValueError:
+                raise ValueError(
+                    "Invalid multiplexed frame received (stream not mapped)"
+                )
+
+            payload = content["payload"]
+
+            # send it on to the application that handles this stream
+            stream_queue.put_nowait(
+                {
+                    "type": "websocket.receive",
+                    "text": await self.encode_json(payload)
+                }
+            )
+            return
+
+        else:
+            raise ValueError("Invalid multiplexed **frame received (no channel/payload key)")
+
+    # Methods to INTERCEPT Stream Applications -> Client
+
+    async def websocket_send(self, message: Dict[str, Any], stream_name: str):
+        """
+        Capture downstream websocket.send messages from the stream applications.
+        """
+        text = message.get("text")
+        # todo what to do on binary!
+
+        json = await self.decode_json(text)
+        data = {
+            "stream": stream_name,
+            "payload": json
+        }
+
+        await self.send_json(data)
+
+    async def websocket_accept(self, message: Dict[str, Any], stream_name: str):
+        """
+        Intercept streams accepting the websocket connection and add them to
+        the set of accepted upstream streams.
+
+        An upstream Application must accept before messages will be
+        forwarded onto it.
+        """
+        self.accepted_upstream_stream.add(stream_name)
+
+    async def websocket_close(self, message: Dict[str, Any], stream_name: str):
+        """
+        Handle the downstream websocket.close message from a stream
+        application.
+        """
+        # just remove the stream from the set of open streams
+        self.accepted_upstream_stream.remove(stream_name)
+        # we don't cancel the task here, that will happen later.
+        # (when the main connection is closed).
+        # the application task could still send messages
+        # if it sends `websocket.accept` it will also start to recived messages
+        # again.
+        await self.close(code=message.get("code", None))
+
+    # De-multiplexing methods
+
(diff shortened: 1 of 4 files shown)


andrewgodwin:
My first reaction to this is that it definitely shouldn't be a mixin - please make it just a normal class! Just reimplement the few bits you need from other classes.

hishnash:
ok sure, so just the `AsyncJsonWebsocketDemultiplexer` class.

channels/generic/websocket.py
```diff
@@ -254,3 +256,155 @@ class AsyncJsonWebsocketConsumer(AsyncWebsocketConsumer):
...
+        super().__init__(scope)
+        self.child_upstream_queues = {}  # type: Dict[str, asyncio.Queue]
+
+    async def __call__(self, receive, send):
+        """
+        Set up child applications and set up self.
+        """
+        child_tasks = []
+        try:
+            self.child_upstream_queues, child_tasks = await self._init_child_applications()
```
andrewgodwin: There's no need to make `init_child_applications` a separate method here really, it just makes it harder to read.

channels/generic/websocket.py
```diff
@@ -254,3 +256,155 @@ class AsyncJsonWebsocketConsumer(AsyncWebsocketConsumer):
...
+                                 lookup: str,
+                                 message: Dict[str, Any]):
+        """
+        Called by child applications when they `base_send(message)`.
+
+        if message.type matches a method of this De-multiplexer this method
+        will be called otherwise message will be passed directly up the
+        application stack.
+        """
+        handler = getattr(self, get_handler_name(message), None)
```
andrewgodwin: Why do you need to intercept it like this? This class is only for websockets, you only need to handle three types of message (accept, send, close).

channels/generic/websocket.py
```diff
@@ -254,3 +256,155 @@ class AsyncJsonWebsocketConsumer(AsyncWebsocketConsumer):
...
+                    "type": "websocket.receive",
+                    "text": await self.encode_json(payload)
+                }
+            )
+            return
+
+        else:
+            raise ValueError("Invalid multiplexed **frame received (no channel/payload key)")
+
+    async def websocket_send(self, message, child_lookup):
```
andrewgodwin: Can we call `child_lookup` -> `stream_name` everywhere? That would make it clearer.

channels/generic/websocket.py
hishnash: I see the use case where users might want to send from the child applications to the de-multiplexer. Since you need to subclass the de-multiplexer to use it having it like this will mean you just need to create the methods you want to call, like any other part of channels.

hishnash:
@andrewgodwin so with respect to `connect` and `disconnect` what i would suggest doing is a bit like v1:

a) the de-multiplexer accepts the connection all on its own. 
https://github.com/django/channels/blob/1.x/channels/generic/websockets.py#L273-L278

b) it calls `connect` to each child once those children accept they are added to the collection of children that can have messages routed to them.

c) if a disconnect comes through all that have responded `accept` will get the message.

c) if they trigger a disconnect they are removed from the set of child_applications and the `asyncio.Task` for them is cancelled.

d) when the last child disconnects (there are no more child applications left open) the de-multiplexer sends the disconnect onwards down.

in v1 the connection was accepted at the de-multiplexer, but if any child app sends a disconnect command it would disconnect the entire de-multiplexer not just disconnect that child from the de-multiplexer.

andrewgodwin:
Hmm, that seems reasonable enough, and people can override and subclass the Multiplexer if they want to put access control/connection denial on that.

ASGI applications should exit their main routine (and thus complete their Future) once they close their connection, so you should wait for all the applications to exit, rather than trying to work out if they have run their `disconnect` function (remember, you only get to send to them, you can't see what they're doing). It's likely worth doing what Daphne does and giving them a 10 second grace period to exit when the socket itself disconnects and they've been sent a disconnect message before just calling `task.cancel()` on them and exiting messily (we should raise a warning if we have to do this). You can see the relevant Daphne code here: https://github.com/django/daphne/blob/f53eb0dda6161c9fca78c8a89a1d94c7b613b1d4/daphne/server.py#L210

hishnash:
ok, given that the de-Multiplexer is already subject to that 10-second timeout should it not be a little less than the Daphne timeout otherwise we will never get the warning from the de-Multiplexer since the de-Multiplexer routine will timeout (10 seconds) before its internal timeout for its children. 

Also if any child raises an exception at any point (that is not StopConsumer) we should kill them, all right?

andrewgodwin:
Hm. You can't rely on daphne's timeout being 10 seconds (it's configurable), so you need to write a try/finally block that makes sure all the children are killed as the multiplexer's main coroutine shuts down, and maybe warn then (though daphne's warning might be enough to let the developer know it's not going well).

The multiplexer's `disconnect` can just run in parallel with the childrens' - there's no longer a socket there to talk to at that point, so it's not like they can do anything to the multiplexer.

If there is an exception, then yes, I think it is fair to drop the entire multiplexer and all children. The alternative is logging the traceback and continuing on, but I think hard failure here is correct.

hishnash:
> The multiplexer's disconnect can just run in parallel with the children's - there's no longer a socket there to talk to at that point, so it's not like they can do anything to the multiplexer.

it might still be nice to have that method be called after since it means all child applications are finished. Maybe you want to do some cleanup etc that needs to wait until all children are finished.

hishnash:
@andrewgodwin this is making `/generic/websocket.py` explode in size I'm going to break this out into a `mutliplexing.py` file.

hishnash:
with the `streams` possibly removing themselves does it make sense to raise a `ValueError` when the `stream` is not listed? does raising a `ValueError` close the socket?

hishnash:
@andrewgodwin I believe I have addressed your points with respect to connections and disconnections if you have time would you be able to give this solution a read over.

in the end it does feel rather convoluted so if you have an idea on how we can exit on Exception (when needed) and wait for the child to close (with a timeout for brute force killing) that does not require all of these extra features observing the child applications for errors.

also, what is the expected behaviour if an exception is raised. (in v1 this did not close the socket) but am i correct that in v2 this will result in the socket being closed? if so then maybe we should not reais exceptions when the `stream_name` cant be found but rather respond with an error frame?

channels/generic/multiplexer.py
```diff
@@ -0,0 +1,322 @@
...
+            #  down (this includes ones that never accepted!)
+            queue.put_nowait(message)
+
+        if self._task_monitors is None:
+            await super().websocket_disconnect(message)
+            return
+
+        try:
+            # wait for all the upstream applications to close down
+            await asyncio.wait_for(
```
andrewgodwin: There's no need to wait for them to be killed. Just block forever and let the outside server kill your whole coroutine if things go bad.

channels/generic/multiplexer.py
```diff
@@ -0,0 +1,322 @@
...
+        finally:
+            try:
+                self._task_monitors.cancel()
+            finally:
+                # this might have raised an exception.
+                # cleanup stream applications
+                for task in stream_tasks.values():
+                    task.cancel()
+
+    async def _monitor_task(self, stream_name: str, task: asyncio.Task):
```
andrewgodwin: StopConsumer should not bubble up out of other consumers (their coroutines will catch it), so I don't know why you have this method? Did you see it happen?

channels/generic/multiplexer.py
hishnash: then we wont be able to raise any warnings at this level we will just get the warning from daphne?

channels/generic/multiplexer.py
hishnash: don't we still need this method so that if just one child application dies (without an exception) we remove it from the set of upstream streams but keep the rest running.  

if we just gather/waiton the coroutines without some wrapper method we will not be able to remove them from the set of open upstream.

hishnash: don't we still need this method so that if just one child application dies (without an exception) we remove it from the set of upstream streams but keep the rest running?  

if we just gather/wait on the coroutines without some wrapper method we will not be able to remove them from the set of open upstream.

hishnash:
> This is still looking way too complicated in general 

I agree, it ran away from itself.

> I'd suggest rewriting it without basing it on a consumer class at all

ok, so what about having a `Demultiplex` class that takes a `multiplexer` class that would be a `JsonWebsocketMultiplexer` in this example (to do the needed web socket specifics). To split out that logic from the raw Demultiplex (handling of upstream applications etc)

channels/generic/multiplexer.py
hishnash: also will this not just result in there always being a warning being raised?

hishnash:
@andrewgodwin I have made a lot of changes, I have tried to break out the logic a bit as well. 

however, I still think on `webseoct.disconect` we need to wait for the child applications to close not just block expecting the server to kill us otherwise we will always get a warning.

I moved all of the Queue observation (and dispatch) to a new parent class that is there to let one monitor both a collection of child application co-routines and the upstream/downstream queues and close them all in the correct manner regardless of who raises an exception.

hishnash:
I completely understand It had run away on itself if you can find a way to manage these child apps without all that complexity that is a much better solution. feel free to close this pull request.

I chose to split out into separate classes just to avoid having a supermassive 500 to 1000 line class to handle all the child task logic.

the big issues I had were:

- ensuring the tasks that wait for the upstream and downstream `Queues` are closed before the run loop is closed. (otherwise, there are loads and loads of warnings raised)
   - this is both the upstream and downstream message `Queues` on the de-multiplexer
   - and the upstream/layer message `Queues` for the child applications.
what would happen is if one child died (eg raised an exception) and I just let the bubble up then the full run-loop would be killed by the server (as it should) but some of the other tasks (as described above) would not be `cancelled` before this.

In addition, I believe it is good practice in this case not to close the downstream connection until all upstream tasks have closed. 

I'm rather new to the `async` `await` world in python so there may well be a much better simpler approach than those that I tried.

hishnash:
Im closing this for know

hishnash:
The changes in https://github.com/django/channels/commit/d6643d96dd59e153cda13681a6d5a8963b9d577c might well have solved the issues I had the resulted in such a complex solution....

Issue #852, opened 2018-02-04
Add Group join/leave functionality to WebsocketConsumer


andrewgodwin:
Previously, the WebsocketConsumer classes had a feature where you could have a `group_names` attribute or property, and it would handle the joining and leaving of groups in connect/disconnect for you. This should come back!

orn688:
Hi, I've been using channels quite a bit recently and would love to help out and take a stab at this. Were you thinking of using the same sort of API as in 1.x with a `connection_groups` method? What was the rationale behind using that on top of the `groups` attribute?

stunaz:
Can this be closed as #901 is merged?

andrewgodwin:
Yup!

PR #901, opened 2018-02-13
Add WebsocketConsumer group join/leave functionality

orn688:
Addresses #852, adding auto group/leave functionality to `WebsocketConsumer` and `AsyncWebsocketConsumer` via the `groups` attribute. The tests will not pass as-is (in fact, they will hang) due to their reliance on `InMemoryChannelLayer`, but they pass with the implementation of `InMemoryChannelLayer` proposed in #863.

### DIFF
--- a/channels/generic/websocket.py
+++ b/channels/generic/websocket.py
@@ -3,7 +3,7 @@ import json
 from asgiref.sync import async_to_sync
 
 from ..consumer import AsyncConsumer, SyncConsumer
-from ..exceptions import AcceptConnection, DenyConnection, StopConsumer
+from ..exceptions import AcceptConnection, DenyConnection, InvalidChannelLayerError, StopConsumer
 
 
 class WebsocketConsumer(SyncConsumer):
@@ -17,9 +17,11 @@ class WebsocketConsumer(SyncConsumer):
         """
         Called when a WebSocket connection is opened.
         """
-        if self.channel_layer:
+        try:
             for group in self.groups:
                 async_to_sync(self.channel_layer.group_add)(group, self.channel_name)
+        except AttributeError:
+            raise InvalidChannelLayerError("BACKEND is unconfigured or doesn't support groups")
         try:
             self.connect()
         except AcceptConnection:
@@ -87,9 +89,11 @@ class WebsocketConsumer(SyncConsumer):
         Called when a WebSocket connection is closed. Base level so you don't
         need to call super() all the time.
         """
-        if self.channel_layer:
+        try:
             for group in self.groups:
                 async_to_sync(self.channel_layer.group_discard)(group, self.channel_name)
+        except AttributeError:
+            raise InvalidChannelLayerError("BACKEND is unconfigured or doesn't support groups")
         self.disconnect(message["code"])
         raise StopConsumer()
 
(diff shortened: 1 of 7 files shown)


orn688:
If this seems alright, I'm happy to add documentation for this feature too.

andrewgodwin:
Yup, this looks good. I just merged that pull request so if you rebase the tests should then pass on Travis.

Documentation would be great to have too - just add to the `consumers.rst` page and that should be enough.

The one minor nitpick I might have is that right now if you specify groups but not a channel layer, there's no hard error - I think it's best to raise an error if you specified groups but there's not actually a channel layer to listen on (right now it will just silently continue)

orn688:
Good point, I'll work on that and add some documentation tomorrow!

channels/generic/websocket.py
```diff
@@ -9,12 +11,15 @@ class WebsocketConsumer(SyncConsumer):
     Base WebSocket consumer. Provides a general encapsulation for the
     WebSocket handling model that other applications can build on.
     """
+    groups = []
```
hishnash: would it not be better to have groups as a `Set[str]`? we do not want to have repeats correct?

channels/generic/websocket.py
```diff
@@ -82,7 +87,9 @@ def websocket_disconnect(self, message):
         Called when a WebSocket connection is closed. Base level so you don't
         need to call super() all the time.
         """
-        # TODO: group leaving
+        if self.channel_layer:
+            for group in self.groups:
+                async_to_sync(self.channel_layer.group_add)(group, self.channel_name)
```
hishnash: `group_add ` -> `group_discard`

orn688: Oops, good catch

channels/generic/websocket.py
andrewgodwin: We're not defining what type it is, the code just wants an iterable provided by the user. No point putting `set()` here rather than `[]`.

andrewgodwin:
Just a heads-up - I don't actively look over individual PR commits as they happen, so let me know in a comment when you want a re-review and I'll take another look through!

orn688:
@andrewgodwin Thanks, should be good to go now!

channels/generic/websocket.py
```diff
@@ -1,20 +1,27 @@
...
     """
+    groups = []
 
     def websocket_connect(self, message):
         """
         Called when a WebSocket connection is opened.
         """
-        # TODO: group joining
+        try:
+            for group in self.groups:
```
hishnash: what about accessing `self. groups` through a method that can be easily overridden (for cases where these groups might be dynamic based on the scope or maybe Mixins that have been added)

eg here call `self._get_groups()`

with the default implementation of '_get_groups`
```python
def _get_groups(self):
    return self.groups
```

starting the method name with `_` ensures that it cant be called through a channel layer message.

@andrewgodwin what do you think?

andrewgodwin: That is what the `@property` decorator is for! It's still easy to override:

```
@property
def groups(self):
    ...
```

orn688:
Do you have any other fixes/suggestions? If not, this is all set on my end!

orn688:
Thanks, no worries :)

Issue #915, opened 2018-02-20
TypeError in Session Middleware


hishnash:
If you're submitting a bug report, please include:

The versions of Channels
==================

(Current Mater) --- ea2fd197 First check the type then check length
python3.6

What you expected to happen vs. what actually happened
===================

Not raise 
> TypeError: string argument without an encoding

when setting a cookie

How you're running Channels (runserver? daphne/runworker? Nginx/Apache in front?)
=========

py.test 

```python
@pytest.mark.django_db(transaction=True)
@pytest.mark.asyncio
async def test_sessions():

    class SimpleHttpApp(AsyncConsumer):
        """
        Barebones HTTP ASGI app for testing.
        """

        async def http_request(self, event):
            self.scope['session'].save()
            assert self.scope["path"] == "/test/"
            assert self.scope["method"] == "GET"
            await self.send({
                "type": "http.response.start",
                "status": 200,
                "headers": [],
            })
            await self.send({
                "type": "http.response.body",
                "body": b"test response",
            })

    communicator = HttpCommunicator(SessionMiddlewareStack(SimpleHttpApp), "GET", "/test/")
    response = await communicator.get_response()
```
- Console logs and full tracebacks of any errors

```
F
tests/test_http.py:275 (test_sessions)
self = <channels.testing.http.HttpCommunicator object at 0x110a524a8>
timeout = 1

    async def receive_output(self, timeout=1):
        """
            Receives a single message from the application, with optional timeout.
            """
        # Make sure there's not an exception to raise from the task
        if self.future.done():
            self.future.result()
        # Wait and receive the message
        try:
            async with async_timeout.timeout(timeout):
>               return await self.output_queue.get()

../../.virtualenvs/channels/lib/python3.6/site-packages/asgiref/testing.py:62: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Queue at 0x110a8c7f0 maxsize=0 _getters[1]>

    @coroutine
    def get(self):
        """Remove and return an item from the queue.
    
            If queue is empty, wait until an item is available.
    
            This method is a coroutine.
            """
        while self.empty():
            getter = self._loop.create_future()
            self._getters.append(getter)
            try:
>               yield from getter
E               concurrent.futures._base.CancelledError

[...]

PR #916, opened 2018-02-20
Fix bug #915 in Session Cookie Middleware

hishnash:
It seems the `bytes(c.output(header=""))` requires a `encoding='utf-8'` if the input is a `str`
see #915


andrewgodwin:
Thanks!

Issue #917, opened 2018-02-20
Auth helper functions [Feature Request]


hishnash:
Detailed description of the overall behaviour
--------------------------------------------

A set of helper functions (like Django's `login(request, user)` `logout(request)` that use the `session` object that is injected into the `scope`

Reasoning why it should be in Channels rather than a third-party app
-----------

Channels does not expose the HttpRequest object to the consumers. It is thus not clear how one might log in a user, (short of creating a mocked HttpRequest object just for the login event). it does, however, include the `AuthMiddleware` that will attempt to expose the currently authenticated user but will not let you log in a user or logout a user.

Examples of usage, if possible (it's easier to discuss concrete code examples)
---------------

```python
class SimpleHttpApp(AsyncConsumer):

    async def http_request(self, event):
       user = # go find a user somewere ;)
 
       login(self.scope, user)

        await self.send({
            "type": "http.response.start",
            "status": 200,
            "headers": [],
        })
       # ...... do some stuff and send that good old body along !


class SimpleLogoutHttpApp(AsyncConsumer):

    async def http_request(self, event):
       logout(self.scope)

        await self.send({
            "type": "http.response.start",
            "status": 200,
            "headers": [],
        })
       # ...... do some stuff and send that good old body along !

```

andrewgodwin:
This will only be possible on `http` connections, not `websocket` connections, and I don't regard it as high-priority as you can do this via views right now. If anyone wants to take it on they are welcome, though!

hishnash:
would there be any way to have Daphne send a message upstream to the Websocket consumer (that listest to the `http.reequest` message) before it responds with the first HTTP response (that upgrades to a WebSocket connection)? for that stuff like auth/rate limiting, etc.

andrewgodwin:
Nothing in the websocket flow goes near the `http` consumers or events, and you can't send things back to the browser as it's not expecting a HTTP connection. The _only_ way you can set cookies is on a genuine HTTP response from a non-websocket HTTP request.

hishnash:
When the browser starts a websocket conneciton it sends at traditional HTTP request 
```http
GET / HTTP/1.1
Connection: Upgrade
Host: 172.20.10.2:8000
Origin: ws://172.20.10.2:8000
Upgrade: websocket
Sec-WebSocket-Version: 13
Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits; server_max_window_bits=15
[...]

PR #923, opened 2018-02-21
Add login, logout for scope based operations (#917)

hishnash:
See issue: #917

added:
* `login(scope, user)`
* `logout(scope)`

updated:
* `get_user(scope)`

this removed the `_cached_user` key from `scope` since if you are logging in and out on another level in the stack we cant assume that modifications to the scope on that level will be acceptable to the scope in other levels so I think it best to always fall back to checking the session.

---

TODO:
- [x] Tests
- [-] CSRF_COOKIE/Token middleware will do this as part of a separate pull request
- [x] Add docs

### DIFF
--- a/channels/auth.py
+++ b/channels/auth.py
@@ -1,16 +1,163 @@
-from django.contrib import auth
-from django.utils.functional import SimpleLazyObject
+from typing import Any, Dict, Optional
 
+from django.conf import settings
+from django.contrib.auth import (
+    BACKEND_SESSION_KEY, HASH_SESSION_KEY, SESSION_KEY, _get_backends, get_user_model, load_backend, user_logged_in,
+    user_logged_out,
+)
+from django.contrib.auth.models import AbstractUser, AnonymousUser
+from django.utils.crypto import constant_time_compare
+from django.utils.functional import SimpleLazyObject
+from django.utils.translation import LANGUAGE_SESSION_KEY
+
+from channels.db import database_sync_to_async
 from channels.sessions import CookieMiddleware, SessionMiddleware
 
 
-def get_user(scope):
-    if "_cached_user" not in scope:
-        # We need to fake a request so the auth code works until we get to
-        # refactor it to take sessions, not requests.
-        fake_request = type("FakeRequest", (object, ), {"session": scope["session"]})
-        scope["_cached_user"] = auth.get_user(fake_request)
-    return scope["_cached_user"]
+def get_user(scope: Dict[str, Any]) -> AbstractUser:
+    """
+    Return the user model instance associated with the given scope.
+    If no user is retrieved, return an instance of `AnonymousUser`.
+    """
+    if "session" not in scope:
+        raise ValueError(
+            "Cannot find session in scope. "
+            "You should wrap your consumer in SessionMiddleware.")
+
+    session = scope["session"]
+
+    user = None
+    try:
+        user_id = _get_user_session_key(session)
+        backend_path = session[BACKEND_SESSION_KEY]
+    except KeyError:
+        pass
+    else:
+        if backend_path in settings.AUTHENTICATION_BACKENDS:
+            backend = load_backend(backend_path)
+            user = backend.get_user(user_id)
+
+            # Verify the session
+            if hasattr(user, "get_session_auth_hash"):
+                session_hash = session.get(HASH_SESSION_KEY)
+                session_hash_verified = session_hash and constant_time_compare(
+                    session_hash,
+                    user.get_session_auth_hash()
+                )
+                if not session_hash_verified:
+                    session.flush()
+                    user = None
+
+    return user or AnonymousUser()
+
+
+def login(scope: Dict[str, Any], user: Optional[AbstractUser], backend=None):
+    """
+    Persist a user id and a backend in the request. This way a user doesn't
+    have to reauthenticate on every request. Note that data set during
+    the anonymous session is retained when the user logs in.
+    """
+
+    if "session" not in scope:
+        raise ValueError(
+            "Cannot find session in scope. "
+            "You should wrap your consumer in SessionMiddleware.")
+
+    session = scope["session"]
+    session_auth_hash = ""
+
+    if user is None:
+        user = scope.get("user", None)
+
+    if user is None:
+        raise ValueError(
+            "User must be passed as an argument or must be "
+            "present in the scope."
+        )
+
+    if hasattr(user, "get_session_auth_hash"):
+        session_auth_hash = user.get_session_auth_hash()
+
+    if SESSION_KEY in session:
+        if _get_user_session_key(session) != user.pk or (
+                session_auth_hash and
+                not constant_time_compare(
+                    session.get(HASH_SESSION_KEY, ""),
+                    session_auth_hash)):
+            # To avoid reusing another user's session, create a new, empty
+            # session if the existing session corresponds to a different
+            # authenticated user.
+            session.flush()
+    else:
+        session.cycle_key()
+
+    try:
+        backend = backend or user.backend
+    except AttributeError:
+        backends = _get_backends(return_tuples=True)
+        if len(backends) == 1:
+            _, backend = backends[0]
+        else:
+            raise ValueError(
+                "You have multiple authentication backends configured and "
+                "therefore must provide the `backend` argument or set the "
+                "`backend` attribute on the user."
+            )
+
+    session[SESSION_KEY] = user._meta.pk.value_to_string(user)
+    session[BACKEND_SESSION_KEY] = backend
+    session[HASH_SESSION_KEY] = session_auth_hash
+    scope["user"] = user
+
+    # note this does not reset the CSRF_COOKIE/Token
+
+    user_logged_in.send(sender=user.__class__, request=None, user=user)
+
+
+def logout(scope: Dict[str, Any]):
+    """
+    Remove the authenticated user's ID from the request and flush their session
+    data.
+    """
+
+    if "session" not in scope:
+        raise ValueError(
+            "Login cannot find session in scope. "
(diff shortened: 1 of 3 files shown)


hishnash:
@andrewgodwin let me know if there are any other changes if not I will do a rebase to compress all the commits.

andrewgodwin:
There's no need to rebase, github has squash merge built in. I'll take a look later.

andrewgodwin:
Hmm, looking over this once more I have noticed the `async_login` and `async_logout` functions - I'd rather just let the user wrap the ones provided as appropriate, so remove them, but mention in the docs that they need to wrap them if calling them from an async context.

hishnash:
@andrewgodwin done but im wanding if I should add `database_sync_to_async` to the call of `get_user` in the `AuthMiddleware` here: https://github.com/hishnash/channels/blob/7cf0ce8f38634b286c7ec863e6a49c50dc6832b4/channels/auth.py#L136

Since it depends on the user's Authentication Backend as it if this is needed.

andrewgodwin:
Hmm, you can't use lazy objects because you might traverse a `database_sync_to_async` call, so everything would have to be done in the function. I am almost tempted to say these should be async functions entirely, to continue with Channels being outwardly async (even if we switch to sync code inside).

hishnash:
so to be clear the AuthMidleware's `__call__` (the same as the __init__ of a Consumer) is called inside a `SyncToAsync` so it would be ok to have an `AsyncToSync`.

hishnash:
just wrap in `database_sync_to_async` were it is likly that some subclass of User/AuthBackend might hit the db:
```python
async def login(scope, user, backend=None):
    """
    Persist a user id and a backend in the request.
    This way a user doesn't have to re-authenticate on every request.
    Note that data set during the anonymous session is retained when the user logs in.
    """
    if "session" not in scope:
        raise ValueError("Cannot find session in scope. You should wrap your consumer in SessionMiddleware.")
    session = scope["session"]
    session_auth_hash = ""
    if user is None:
        user = scope.get("user", None)
    if user is None:
        raise ValueError("User must be passed as an argument or must be present in the scope.")
    if hasattr(user, "get_session_auth_hash"):
        session_auth_hash = await database_sync_to_async(user.get_session_auth_hash)()
    if SESSION_KEY in session:
        if await database_sync_to_async(_get_user_session_key)(session) != user.pk or (
                session_auth_hash and not
                constant_time_compare(session.get(HASH_SESSION_KEY, ""), session_auth_hash)):
            # To avoid reusing another user's session, create a new, empty
            # session if the existing session corresponds to a different
            # authenticated user.
            await database_sync_to_async(session.flush)()
    else:
        await database_sync_to_async(session.cycle_key)()
    try:
        backend = backend or user.backend
    except AttributeError:
        backends = await database_sync_to_async(_get_backends)(return_tuples=True)
        if len(backends) == 1:
            _, backend = backends[0]
        else:
            raise ValueError(
                "You have multiple authentication backends configured and therefore must provide the `backend` "
                "argument or set the `backend` attribute on the user."
            )
    session[SESSION_KEY] = await database_sync_to_async(user._meta.pk.value_to_string)(user)
    session[BACKEND_SESSION_KEY] = backend
    session[HASH_SESSION_KEY] = session_auth_hash
    scope["user"] = user
    # note this does not reset the CSRF_COOKIE/Token
    await database_sync_to_async(user_logged_in.send)(sender=user.__class__, request=None, user=user)
```

or wrap the entier thing
```python
async def login(scope, user, backend=None):
    @database_sync_to_async
    def _login(.....):
          ....
    return await _login(......)
```

hishnash:
@andrewgodwin I switched to `async` methods so you can see how it feels (can revert back if you feel you prefer the sync in the end)

andrewgodwin:
Hmm, I really don't know which one is better at this point. Give me the weekend to think about it.

channels/auth.py
```diff
@@ -28,7 +137,7 @@ def __call__(self, scope):
             raise ValueError("AuthMiddleware cannot find session in scope. SessionMiddleware must be above it.")
         # Add it to the scope if it's not there already
         if "user" not in scope:
-            scope["user"] = SimpleLazyObject(lambda: get_user(scope))
+            scope["user"] = async_to_sync(get_user)(scope)
```
andrewgodwin: You can't do this here, as you'd be doing a blocking operation in the ASGI app's first callable (which can't block). I'm surprised this didn't raise an error during testing... still, you should keep the lambda for now (that means you can still accidentally call it during an async consumer, but I'd like to fix that separately)

docs/topics/sessions.rst
```diff
@@ -75,3 +75,52 @@ connections or HTTP views won't be able to see the changes.
     If you are in a long-polling HTTP consumer, you might want to save changes
     to the session before you send a response. If you want to do this,
     call ``scope["session"].save()``.
+
+
+How to log a user in/out
+------------------------
+
+Within your consumer you can call ``login(scope, user, backend=None)`` to login a user.
```
andrewgodwin: call -> await
"to login a user" -> "to log a user in"

docs/topics/sessions.rst
```diff
@@ -75,3 +75,52 @@ connections or HTTP views won't be able to see the changes.
...
+This requires that your scope has a ``session`` object the best way to do this is to
```
andrewgodwin: "object the best way" -> "object; the best way"

docs/topics/sessions.rst
```diff
@@ -75,3 +75,52 @@ connections or HTTP views won't be able to see the changes.
...
+ensure your consumer is wrapped in a ``SessionMiddlewareStack`` or a ``AuthMiddlewareStack``.
+
+You can logout a user with the ``logout(scope)`` function.
```
andrewgodwin: "function" -> "async function"

docs/topics/sessions.rst
```diff
@@ -75,3 +75,52 @@ connections or HTTP views won't be able to see the changes.
...
+
+If you are in a WebSocket consumer, or logging-in after the first response has been sent in a http consumer,
+the session is populated **but will never be saved automatically** - you must call ``scope["session"].save()``
+after login in your consumer code::
+
```
andrewgodwin: Remove extra newline

docs/topics/sessions.rst
```diff
@@ -75,3 +75,52 @@ connections or HTTP views won't be able to see the changes.
...
+    class ChatConsumer(AsyncWebsocketConsumer):
+
+        ...
+
+        async def receive(self, text_data):
+            ...
+            # login the user to this session.
+            await login(self.scope, user)
+            # save the session (if the session backend does not access the db you an use `sync_to_async`)
```
andrewgodwin: "you an use" -> "you can use"

docs/topics/sessions.rst
```diff
@@ -75,3 +75,52 @@ connections or HTTP views won't be able to see the changes.
...
+            await database_sync_to_async(self.scope["session"].save)()
+
+When calling ``login(scope, user)``, ``logout(scope)`` or ``get_user(scope)`` from an sync function you will need to
+wrap them in ``async_to_sync``::
+
```
andrewgodwin: Remove extra newline

docs/topics/sessions.rst
```diff
@@ -75,3 +75,52 @@ connections or HTTP views won't be able to see the changes.
...
+        def receive(self, text_data):
+            ...
+            async_to_sync(login)(self.scope, user)
+            self.scope["session"].save()
+
+.. note::
+
+    If you are using a long running consumer, websocket or long-polling HTTP it is possible
+    the user has been logged out of their session elsewere.
```
andrewgodwin: elsewere -> elsewhere

docs/topics/sessions.rst
```diff
@@ -75,3 +75,52 @@ connections or HTTP views won't be able to see the changes.
...
+    Use ``get_user(scope)`` to be sure that the user is still logged in.
```
andrewgodwin: "Use" -> "You can periodically use"

tests/security/test_auth.py
```diff
@@ -0,0 +1,198 @@
...
+    assert await get_user(scope) == user
+
+
+@pytest.mark.django_db(transaction=True)
+@pytest.mark.asyncio
+async def test_login_no_session_in_scope():
+
+    with pytest.raises(
+            ValueError,
+            match="Cannot find session in scope. You should wrap your consumer in SessionMiddleware."):
```
andrewgodwin: Move closing bracket to next line

tests/security/test_auth.py
```diff
@@ -0,0 +1,198 @@
...
+    assert BACKEND_SESSION_KEY in session
+    assert HASH_SESSION_KEY in session
+
+    assert isinstance(await get_user(scope), await database_sync_to_async(get_user_model)())
+    assert await get_user(scope) == user
+
+
+@pytest.mark.django_db(transaction=True)
+@pytest.mark.asyncio
+async def test_login_no_session_in_scope():
```
andrewgodwin: Docstrings on each test please saying what it is testing as an English sentence

tests/security/test_auth.py
```diff
@@ -0,0 +1,198 @@
...
+    BACKEND_SESSION_KEY, HASH_SESSION_KEY, SESSION_KEY, get_user_model, user_logged_in, user_logged_out,
+)
+from django.contrib.auth.models import AnonymousUser
+
+from asgiref.sync import sync_to_async
+from channels.auth import get_user, login, logout
+from channels.db import database_sync_to_async
+
+
+class catch_signal():
```
andrewgodwin: Docstring saying what this does please

hishnash:
@andrewgodwin let me know if there are any other changes?

hishnash:
@andrewgodwin ... should be good now 🤞

andrewgodwin:
Great, thanks!

Issue #918, opened 2018-02-20
WebsocketCommunicator does not split out query string


bruecksen:
At the moment I'm trying to fix my test cases in channels 2 and found out, that the `WebsocketComunicator` doesn't consider any query parameters. Is it somehow possible to pass a query parameter into the scope? I implemented a token based authentication middleware which is extracting the token from a query string. 

Testing a custom authentication middleware should be possible in Channels. At the moment it is not  possible to write a test case for the custom authentication example from the documentation: 
http://channels.readthedocs.io/en/latest/topics/authentication.html#custom-authentication

I'm using:
```
channels==2.0.2
channels-redis==2.0.3
daphne==2.0.3
```

Thanks for any help or advice.

hoefling:
I faced the same issue lately when started moving the code to using `channels==2.0` and here is the workaround I'm using right now:

```python
from channels.testing import WebsocketCommunicator as WebsocketCommunicatorOrig

class WebsocketCommunicator(WebsocketCommunicatorOrig):
    """
    The original websocket communicator does not allow to pass query string to scope, so
    we write our own communicator that allows us to do exactly that (``query`` arg in constructor).
    """

    def __init__(self, application: AsyncConsumer, path: str, headers: Optional[List[str]] = None,
                 subprotocols: Optional[List[str]] = None, query: Optional[str] = None) -> None:
        self.scope = {
            'type': 'websocket',
            'path': path,
            'headers': headers or [],
            'subprotocols': subprotocols or [],
            'query_string': query or '',
        }
        # avoid calling super() here or the query string will be lost
        ApplicationCommunicator.__init__(self, application, self.scope)
```

This is pretty ugly since you have to avoid calling the `channels.testing.WebsocketCommunicator`'s init. Modifying the scope after the `super` call will have no effect and there's no way to pass custom args to the scope in `WebsocketCommunicator`'s init. But as long as the `WebsocketCommunicator` does nothing besides filling the scope, this should work.

Ideally, I'd see the `channels.testing.WebsocketCommunicator` accepting keyword args for the scope so one can pass whatever he needs to mock the switching of protocols, i.e.

```python
class WebsocketCommunicator(ApplicationCommunicator):

    def __init__(self, application, path, **kwargs):
        headers = kwargs.pop('headers', [])
        subprotocols = kwargs.pop('subprotocols', [])
        self.scope = {
            "type": "websocket",
            "path": path,
            "headers": headers,
            "subprotocols": subprotocols,
            **kwargs
[...]

PR #932, opened 2018-02-25
Split path string to path and query strings in WebsocketCommunicator

hoefling:
Fix proposal for #918. Thanks in advance for the review!

### DIFF
--- a/channels/testing/websocket.py
+++ b/channels/testing/websocket.py
@@ -13,19 +13,13 @@ class WebsocketCommunicator(ApplicationCommunicator):
     """
 
     def __init__(self, application, path, headers=None, subprotocols=None):
+        if not isinstance(path, str):
+            raise TypeError("Expected str, got {}".format(type(path)))
         parsed = urlparse(path)
-        query_string = parsed.query
-        # ASGI spec: query_string should be a byte string
-        if isinstance(query_string, str):
-            query_string = query_string.encode("utf-8")
-        path_string = parsed.path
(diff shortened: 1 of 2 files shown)


andrewgodwin:
One small niggle - the `query_string` has to be a non-url-decoded bytestring while the `path` has to be a url-decoded unicode string (see the spec here: https://github.com/django/asgiref/blob/master/specs/www.rst), so you need to get those aligned here.

I'd also love to see a test that verifies that!

hoefling:
Sorry for the long delay! Please review the changes, especially if you like the added test. I tried to create a suggestion for testing against the ASGI spec - the `test_connection_scope` is rather primitive, but could be extended to cover the whole spec for the websocket connection scope by adding more validators for each scope field. On the other hand, should you encounter some exotic paths you'd like to check the scope for, just add them to the `paths` list and they will be included in the spec tests.

hoefling:
Or maybe this is an overkill for the simple task of testing the `WebsocketCommunicator`?

andrewgodwin:
Two things:

- I'd prefer to not auto-encode/decode things to `utf8`, but instead to straight up raise errors if the input does not match the required type.
- Just combine the three test communicators into one, IMO. No need to parameterise them like that.

hoefling:
Yeah, I don't like that idea either now. Please review the latest changes!

andrewgodwin:
Looks great, thanks!

PR #935, opened 2018-02-26
Add de-multiplexer

hishnash:
Adds `AsyncJsonWebsocketDemultiplexer`

@andrewgodwin  based on your changes in https://github.com/django/channels/commit/d6643d96dd59e153cda13681a6d5a8963b9d577c I was able to simplify this a lot :) without Futures/Tasks being raising lots of warnings.

key features of `AsyncJsonWebsocketDemultiplexer`

* will send `websocket.accept` back only after the first upstream application responds in kind.
* will block upstream `message frames` from being sent to upstream applications until said upstream application response with `websocket.accept`
* if an upstream application sends `websocket.close` will remove it from the set up upstream applications that can receive `message frames`
* if the last open upsteam application sends `websocket.close` then it will propagate that down to the server so that the connection closes.
* when a `websocket.disconect` is received this is sent to all the child applications so they can close. They have `application_close_timeout=5` to close before they will be killed by the parent.
* if the consumer loop of an upstream application completes then all upstream applications will be closed as well as the de-multiplexer!.

# TODO
- [ ] handle what happens if the upstream applications sends a binary frame?
- [ ] Add lots of docs

### DIFF
--- a/channels/generic/demultiplexer.py
+++ b/channels/generic/demultiplexer.py
@@ -1,0 +1,200 @@
+import asyncio
+from functools import partial
+
+from channels.consumer import get_handler_name
+from channels.generic.websocket import AsyncJsonWebsocketConsumer
+
+
+class AsyncJsonWebsocketDemultiplexer(AsyncJsonWebsocketConsumer):
+    """
+    JSON-understanding WebSocket consumer subclass that handles de-multiplexing streams using a "stream" key in a
+    top-level dict and the actual payload in a sub-dict called "payload".
+    This lets you run multiple streams over a single WebSocket connection in a standardised way.
+    Incoming messages on streams are dispatched to consumers so you can just tie in consumers the normal way.
+    """
+    applications = {}
+    application_close_timeout = 5
+
+    def __init__(self, scope):
+        super().__init__(scope)
+        self.application_streams = {}
+        self.application_futures = {}
+        self.applications_accepting_frames = set()
+        self.closing = False
+
+    async def __call__(self, receive, send):
+        loop = asyncio.get_event_loop()
+        # create the child applications
+        await loop.create_task(self._create_upstream_applications())
+        # start observing for messages
+        message_consumer = loop.create_task(super().__call__(receive, send))
+        try:
+            # wait for either an upstream application to close or the message consumer loop.
+            await asyncio.wait(
+                list(self.application_futures.values()) + [message_consumer],
+                return_when=asyncio.FIRST_COMPLETED
+            )
+        finally:
+            # make sure we clean up the message consumer loop
+            message_consumer.cancel()
+            try:
+                # check if there were any exceptions raised
+                await message_consumer
+            except asyncio.CancelledError:
+                pass
+            finally:
+                # Make sure we clean up upstream applications on exit
+                for future in self.application_futures.values():
+                    future.cancel()
+                    try:
+                        # check for exceptions
+                        await future
+                    except asyncio.CancelledError:
+                        pass
+
+    async def _create_upstream_applications(self):
+        """
+        Create the upstream applications.
+        """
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Oh, I have a half-written multiplexer locally already, sorry, I thought I said I would work on it on the last PR - I'll put it up into a branch at some point so you can see, but it has different behaviour to your design (like no need for timeouts, different accept flow, keeping alive when a child application completes).

hishnash:
> Oh, I have a half-written multiplexer locally already, sorry,

no problem just wanted to share this version given the changes that were made to `await_many_dispatch` is know a lot cleaner (compared to before)

> keeping alive when a child application completes

I'm really not sure what should be done with this. In v1 code if a child app closed then it would close the entire de-multiplexer this is simpler for sure but might have some side effects for the other sibling applications. 
But then again it does remove some strangeness for the client in that the set of `steam` names that are available is stable so the connection is either open (and accepts on all steams) or closed and accepts on none rather than open but the set of acceptable steams is constantly changing.

andrewgodwin:
Yeah, I don't want any of the child applications to affect any others - so if one exits or closes, it just stops being forwarded messages.

andrewgodwin:
Pushed up the code I'm working on here, but it's not complete yet: https://github.com/django/channels/compare/multiplexing

hishnash:
Nice, I like the buffer idea :)

at least as a read over it it does not look like if an exception is raised in say the second child application it will not be captured? 

What do you think should happen if a child raises an exception? I assumed we would want to bubble that up and kill everything.

hishnash:
as a general question moving from v2 to v3 I am correct that in v2 if an exception was raised the connection was not always lost but in v3 it is? since before Daphne and the consumers were separate processes so a different worker would just pick up the next incoming message.

With this the case I think it would be nice to have a clean way to override the behaviour of the de-multiplexer when an upstream steam cant is found, maybe you don't want to crash the connection but instead.

andrewgodwin:
I'm not sure what you mean by v3, but I also don't want to make it too customisable as that usually makes it harder to maintain and test.

hishnash:
> I'm not sure what you mean by v3

sorry mixed up version numbers.

in channels v1 if a consumer raised an exception it did not mean the connection was closed? but in v2 it does no?

> but I also don't want to make it too customisable as that usually makes it harder to maintain and test.

👍

andrewgodwin:
I can't even remember what channels v1 did, and I'm not trying to match its disconnection behaviour, as multiplexing there was very limited and buggy.

hishnash:
ok so the key features your going for here are:

* frames are not sent upstream to child apps until they `websocket.accept` they are rather buffered 👍 
* when an upstream app sends  `websocket.accept` downstream any buffered frames are pushed up and new frames are permitted to flow freely 👍 
* if a child app dies make sure that exception is seen by the developer but do not kill/affect any of its sibling apps 👍 

* if an upstream child app sends `websocket.close` just top sending it new frames (delete its queue and buffer reference) .. this means it will not get `websocket.disconnect` as it would normally if it were not behind a de-multiplexer. (maybe the de-multiplexer should emulate a client disconnection and send `websocket.disconnect` back up to the child app that sent `websocket.close` so that the normal `disconnect` method etc can be handled.)

* if a message is sent to a `stream_name` that is not mapped raise an exception and drop the entire WebSocket connection. 

so the de-multiplexer consumer (and all the children unless they raise exceptions) will run forever until the client closes the connection. even if all child apps send `websocket.close`?

pashinin:
Can I use this PR? As I understand demultiplexer is already working except some minor bugs and it will not change dramatically.

andrewgodwin:
@pashinin The version I'll be working towards is https://github.com/django/channels/compare/multiplexing, not this PR - I'm going to close this for now to make that clearer.

Issue #841, opened 2018-02-02
Implement HTTP generic consumers


andrewgodwin:
The HTTP event spec in ASGI is a little complex, and it would be nice to have an Async-capable generic consumer for HTTP to allow easier writing of long-poll applications. Something like:

```
class LongPoller(AsyncHttpConsumer):
    async def request(self, body):
        await asyncio.sleep(10)
        await self.send_response(200, "foo")
```

matthiask:
That's what I have for now. Do you think it would make sense to reuse `AsgiRequest` here?

```
class AsyncHttpConsumer(AsyncConsumer):
    async def __call__(self, receive, send):
        self.send = send
        body = []
        while True:
            message = await receive()
            if message['type'] == 'http.disconnect':
                return
            else:
                if 'body' in message:
                    body.append(message['body'])
                if not message.get('more_body'):
                    # req = AsgiRequest(self.scope, b''.join(body))
                    await self.handle(b''.join(body))
                    return

    async def send_response(self, body, *, headers=[], status=200):
        await self.send({
            'type': 'http.response.start',
            'status': status,
            'headers': headers,
        })

        await self.send({
            'type': 'http.response.body',
            # Probably not a good place for the encode call, but works for now...
            'body': body.encode('utf-8'),
        })
```

I'd be happy to contribute something, but would need some guidance.

hishnash:
would it not be nice to be able to start the `send_response` as soon as all the headers (the first HTTP message) have been received. 

> An HTTP/1.1 (or later) client sending a message-body SHOULD monitor the network connection for an error status while it is transmitting the request. If the client sees an error status, it SHOULD immediately cease transmitting the body.

the server might want to reject the request before you prosses the full body (for example you might have a file size limit) or just want to reject an upload based on some auth faster so that the client does not need to wait to fully upload. Or if you are uploading a file (and include the md5 in the headers) the server could respond immediately if the file is already uploaded.

andrewgodwin:
I think reusing `AsgiRequest` would be a nice feature if it's possible, but it does a lot of stuff internally (like reading the HTTP body) that would probably need to be disabled if you did that.

Otherwise, I think it might be best to have optional `send_headers` and `send_body` methods that you can call individually if you want, as well as an overall `send_response`. I would also double-check how Django handles encoding unicode string bodies and replicate that (I don't think it just always makes them `utf8` without putting a header on at least)

matthiask:
@andrewgodwin Thanks!

[...]

Issue #524, opened 2017-02-10
Document long polling


Christophe31:
It looks like channel handle long polling via ResponseLater exception but no test or documentation on how to retreive connexion to send data back and close it.

I supose we may use channels.handler.ViewConsumer the same way we would use a WebsocketConsumer, but I didn't find this info in any doc.

ivorbosloper:
I'm testing ResponseLater for one of my views but can't make it work. I have some middleware, and it seems that any middleware will always force a HttpResponse. The ResponseLater exception won't reach the Channels code (in AsgiHandler) ignoring this specific Exception, it will get a HttpResponse object with HTML content describing the error.

The AsgiHandler wraps BaseHandler.get_response() in a try/except and then ignores the ResponseLater exception. But if my view raises the ResponseLater, it is converted to a HttpResponse before it reaches the AsgiHandler. I'm still using MIDDLEWARE_CLASSES, but I'm not sure if this is the reason.

The channels/handler.AsgiHandler relies uses BaseHandler.load_middleware(), which wraps a `convert_exception_to_response` around all middleware and around get_response. If this is not somehow monkeypatched the exception won't be bubbled up.

In channels < 0.8 the ResponseLater exception thrown by a Django View was supported through monkeypatching in channels/hacks.py . This was removed in commit #b9464ca149a4f66ebeb8801254cdd6d1b12cd58d . Restoring it doesn't help me either.

Is ResponseLater still supported?

ivorbosloper:
Another design could be to annotate the Response object with an attribute which signifies the ResponseLater intent. Like:

```
def my_view(request):
    response = HttpResponse("", status=200)
    response._channels_response_later = True
    return response
```

The channels/handler.AsgiHandler should then check for this attribute. This will work as long as the middleware does not replace the response object, but at least it's a working option.

andrewgodwin:
Yes, ResponseLater does not really interact well with any middleware. It's been replaced in Channels 2 by async HTTP views instead, so you can `await`, but of course you give up old-style middleware entirely (channels 2 allows for ASGI middleware, which I need to document, and which works quite like current Django middleware but allows for async stuff)

ivorbosloper:
@andrewgodwin Thanks for your reply. I suspect Channels 2 async requires python3. I would love to have _some_ mechanism supported with middleware to skip responses, at least until we all can move to Channels 2. I think status code 100 (Continue) or some custom header would be more suitable for decorating the response. I've created pull-request #803 

[...]

PR #937, opened 2018-02-28
AsyncHttpConsumer with docs for long polling/SSE

matthiask:
Refs #841, #524 

This is a very rough initial draft for an asynchronous HTTP consumer. 

I was torn a bit whether I should even add some code. For me, the documentation part is much more interesting and important for now. There certainly are ways to make the functionality more developer-friendly later, so maybe we do not want to decide what the recommended API should be yet.

Also, no synchronous version yet because going completely async is in some horrible way easier to understand for me  😊

TODO:

- [x] Tests
- [x] Default content type / charset

### DIFF
--- a/channels/generic/http.py
+++ b/channels/generic/http.py
@@ -1,3 +1,5 @@
+from django.utils import six
+
 from channels.consumer import AsyncConsumer
 
 
@@ -47,16 +49,16 @@ class AsyncHttpConsumer(AsyncConsumer):
 
     async def send_body(self, body, *, more_body=False):
         """
-        Sends a response body to the client. The method expects a unicode
-        string.
+        Sends a response body to the client. The method expects a bytestring.
 
         Set ``more_body=True`` if you want to send more body content later.
         The default behavior closes the response, and further messages on
         the channel will be ignored.
         """
+        assert isinstance(body, six.binary_type), "Body is not bytes"
         await self.send({
             "type": "http.response.body",
-            "body": body.encode("utf-8"),
+            "body": body,
             "more_body": more_body,
         })
 
--- a/channels/generic/http.py
+++ b/channels/generic/http.py
@@ -1,5 +1,3 @@
-from django.utils import six
-
 from channels.consumer import AsyncConsumer
 
 
@@ -55,7 +53,7 @@ class AsyncHttpConsumer(AsyncConsumer):
         The default behavior closes the response, and further messages on
         the channel will be ignored.
         """
-        assert isinstance(body, six.binary_type), "Body is not bytes"
+        assert isinstance(body, bytes), "Body is not bytes"
         await self.send({
             "type": "http.response.body",
             "body": body,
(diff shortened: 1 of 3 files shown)


andrewgodwin:
Yup, that seems pretty reasonable. 

Things I think could be done:
* Accept both dicts and list of tuples for headers (don't see a reason why not here)
* Don't use AsgiRequest at all for now, sessions and auth should already be handled by the ASGI versions of their middleware anyway
* Docstrings everywhere

I don't think a sync version is needed, because if you want that, well, you have Django views!

matthiask:
I've added tests and docs and squashed the history into a single commit (no need for fixup commits fixing previous fixups). Some feedback would be awesome if you find the time. Thanks :-)

channels/generic/http.py
```diff
@@ -0,0 +1,82 @@
...
+
+    async def send_body(self, body, *, more_body=False):
+        """
+        Sends a response body to the client. The method expects a bytestring.
+
+        Set ``more_body=True`` if you want to send more body content later.
+        The default behavior closes the response, and further messages on
+        the channel will be ignored.
+        """
+        assert isinstance(body, six.binary_type), "Body is not bytes"
```
andrewgodwin: No need to use `six` - Channels 3 is Python 3 only!

docs/topics/consumers.rst
```diff
@@ -277,3 +281,66 @@ AsyncJsonWebsocketConsumer
...
+            # Set "more_body" to tell the interface server to not
+            # finish the response yet:
+            await self.send_body(b"", more_body=True)
+
+        async def chat_message(self, event):
+            # Send JSON and finish the response:
+            await self.send_body(json.dumps(event).encode("utf-8"))
+
+Of course you can also use those primitives to implement a HTTP endpoint for
+`Server-sent events`_::
```
andrewgodwin: I'd prefer this as an inline link (I know, I'm weird, but the rest of the docs work like that and I like consistency)

matthiask:
👍 Thanks! 

I was thrown off by other uses of `six` in Channels.

andrewgodwin:
Alright, merged, thanks for this!

I didn't realise I had left some `six` in there, that's be missing them during the 1-to-2 conversion. I will go kill those now.

erm:
I found this in the docs https://channels.readthedocs.io/en/latest/topics/consumers.html#asynchttpconsumer, but it hasn't been included in a new release/made available on PyPI as of yet - how are new releases handled currently?

andrewgodwin:
They happen when I have the time - this next release is a major one, and work has been very busy recently, so it might be another week or two. You can use the master branch if you want to play with it, though - like Django core, master is usually quite stable.

alega19:
@andrewgodwin 
> I don't think a sync version is needed, because if you want that, well, you have Django views!

Why? How to implement Long polling with views? It needs a consumer. AsyncHttpConsumer is good until no need to call sync functions (orm, etc..).

Issue #929, opened 2018-02-24
Fix and improve OriginValidator


bluetech:
> Your OS and runtime environment, and browser if applicable

Linux, Firefox.

> The versions of Channels, Daphne, Django, Twisted, and your ASGI backend (channels_redis normally)

- Channels 2.0.2 
- Django 2.0.2

> How you're running Channels (runserver? daphne/runworker? Nginx/Apache in front?)

All of the above.

> What you expected to happen vs. what actually happened

Regarding this example https://github.com/django/channels/blob/d6643d96dd59e153cda13681a6d5a8963b9d577c/docs/topics/security.rst#websockets

- The correct syntax for subdomain wildcard is `.goodsite.com` not `*.goodsite.com` ([reference](https://docs.djangoproject.com/en/2.0/ref/settings/#allowed-hosts)).

- `.goodsite.com` includes `goodsite.com` so `goodsite.com` is redundant.

- It should perhaps be mentioned that `*` is a special value that allows all.

---

Regarding `OriginValidator` itself, why does it only validate the domain rather than the full origin? (An [Origin](https://tools.ietf.org/html/rfc6454) is `scheme://domain:port`).

In terms of the web security model, the following are all different origins:

```
http://example.com:80
http://example.com:8000
https://example.com:443
https://example.com:8443
```

and it should be possible to allow any subset of those rather than all or none.

I know that Django's `ALLOWED_HOSTS` disregards the port (a [Host](https://tools.ietf.org/html/rfc7230#section-5.4) is `domain:port`, no scheme). I don't really agree with that decision, however I don't think it applies to Origins as well: `ALLOWED_HOSTS` can be seen as just an additional protection layer over what the webserver should be doing already; but the list of allowed origins is an explicit policy decision unrelated to the webserver, namely, which 3rd party web sites I want to allow to connect to my WebSocket server.

andrewgodwin:
The wildcard matching is a good point - we should update the code to match Django's implementation, rather than keeping this apparently-broken one from Channels 1. That's going to have to be done in a backwards-compatible way now, though, which I think is fine.

The problem of including the port is a good one. I think the right thing to do is add a mode where it doesn't care about the port for the one that reads from the settings file, and let you be more specific manually if you want.

The scheme should also be something you can probably determine on (to only allow connections from SSL pages), but again optionally, in this case if you provide it we enforce it.

andrewgodwin:
Was fixed in #940

PR #940, opened 2018-02-28
Fix and improve OriginValidator

Chukak:
See issue  #929 

Added: 
* methods
    * `self.valid_origin_full(self, origin)` 
    * `self.validate_host_full(self, origin)`
    * `self.is_allowed_origin(self, origin, pattern)`
    * `self.get_origin_ports(self, origin)`
    * `self.check_pattern(self, origin, pattern)`
* attributes
    * `self.full`
    * `self.check_cert`

The ability to check only domains (by default) or to check the entire source is added. You can add both complete and incomplete sources to the list of allowed domains, but for complete verification, at least a schema and domain are required. You can also check the ssl certificate.

Also included are tests and an update for the documentation.

### DIFF
--- a/channels/security/websocket.py
+++ b/channels/security/websocket.py
@@ -1,7 +1,9 @@
+import ssl
+from http import client as http_client
 from urllib.parse import urlparse
 
 from django.conf import settings
-from django.http.request import validate_host
+from django.http.request import is_same_domain, validate_host
 
 from ..generic.websocket import AsyncWebsocketConsumer
 
@@ -10,26 +12,38 @@ class OriginValidator:
     """
     Validates that the incoming connection has an Origin header that
     is in an allowed list.
+
+    If full attr is True, any origin from allowed_origins must be scheme://hostname[:port].
+    Port is optional, but recommended.
+    Check_cert works only with full attr.
     """
 
-    def __init__(self, application, allowed_origins):
+    def __init__(self, application, allowed_origins, full=False, check_cert=False):
         self.application = application
         self.allowed_origins = allowed_origins
+        # Check state of variable full
+        self.full = full
+        # Check certificate or not. Works only if full True
+        self.check_cert = check_cert if full else False
 
     def __call__(self, scope):
         # Make sure the scope is of type websocket
         if scope["type"] != "websocket":
             raise ValueError("You cannot use OriginValidator on a non-WebSocket connection")
         # Extract the Origin header
-        origin_host = None
+        origin_hostname = None
+        result_parse = None
         for header_name, header_value in scope.get("headers", []):
             if header_name == b"origin":
                 try:
-                    origin_host = urlparse(header_value.decode("ascii")).hostname
+                    # Set ResultParse and origin hostname
+                    result_parse = urlparse(header_value.decode("ascii"))
+                    origin_hostname = result_parse.hostname
                 except UnicodeDecodeError:
                     pass
-        # Check to see if the origin header is valid
-        if self.valid_origin(origin_host):
+        # Check to see if the origin header is valid. Full or default check
+        valid = self.valid_origin(origin_hostname) if not self.full else self.valid_origin_full(result_parse)
+        if valid:
             # Pass control to the application
             return self.application(scope)
         else:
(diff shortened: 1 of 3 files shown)


bluetech:
By an unfortunate coincidence, I have just written [my own version](https://github.com/bluetech/channels/commits/fix-origin-validator) of this change. Since you already submitted, I'll tuck mine away, though you can take a look and see if there's something helpful there.

Thanks for working on the issue!

channels/security/websocket.py
```diff
@@ -10,26 +12,38 @@ class OriginValidator:
...
     Validates that the incoming connection has an Origin header that
     is in an allowed list.
+
+    If full attr is True, any origin from allowed_origins must be scheme://hostname[:port].
+    Port is optional, but recommended.
+    Check_cert works only with full attr.
     """
 
-    def __init__(self, application, allowed_origins):
+    def __init__(self, application, allowed_origins, full=False, check_cert=False):
```
andrewgodwin: Can we have a better name for this than `full`?

channels/security/websocket.py
```diff
@@ -40,9 +54,88 @@ def valid_origin(self, origin):
         # None is not allowed
         if origin is None:
             return False
+        # Get only hostname all allowed origins
+        # Only if not full: https://domain.example.com:8443 -> domain.example.com
+        #   or //domain.example.com - > domain.example.com
+        #   or .example.com -> .example.com
+        self.allowed_origins = [urlparse(pattern).hostname or urlparse("//" + pattern).hostname
```
andrewgodwin: Can you rewrite this without using `or`? Two in a row makes it harder to skim over.

channels/security/websocket.py
```diff
@@ -40,9 +54,88 @@ def valid_origin(self, origin):
...
+                                or pattern for pattern in self.allowed_origins]
```
andrewgodwin: The current code indent style is one level and the trailing bracket on the next line - please try and match that!

channels/security/websocket.py
```diff
@@ -10,26 +12,38 @@ class OriginValidator:
...
-                    origin_host = urlparse(header_value.decode("ascii")).hostname
+                    # Set ResultParse and origin hostname
+                    result_parse = urlparse(header_value.decode("ascii"))
+                    origin_hostname = result_parse.hostname
                 except UnicodeDecodeError:
                     pass
-        # Check to see if the origin header is valid
-        if self.valid_origin(origin_host):
+        # Check to see if the origin header is valid. Full or default check
+        valid = self.valid_origin(origin_hostname) if not self.full else self.valid_origin_full(result_parse)
```
andrewgodwin: I'd rather do this check inside a single `valid_origin` function.

channels/security/websocket.py
```diff
@@ -40,9 +54,88 @@ def valid_origin(self, origin):
...
+        return False
+
+    def get_origin_ports(self, origin):
+        if origin.port is not None:
+            # Return tuple (origin.port, )
+            return origin.port,
+        # if origin.port doesn`t exists
+        if origin.scheme == "http":
+            # Default ports return for http
+            return 80, 8080
```
andrewgodwin: Is port 8080 actually specified as a default HTTP port anywhere?

hundredeir: Yes

channels/security/websocket.py
Chukak: Can you suggest a better name than `` full``, maybe ``origin`` or ``scheme``. Is it necessary to leave the possibility of verifying the certificate or include it in the full check origin.

Chukak:
I updated my code. Now, ``OriginValidator`` does not require any options. In ``allowed_origins``, you can pass only domain or origin as scheme://domain[:port] (for exapmple,  ``.domain.com``, ``*``, ``http://.domain.com``.)

channels/security/websocket.py
```diff
@@ -21,27 +24,134 @@ def __call__(self, scope):
...
+    def validate_origin(self, parsed_origin):
+        """
+        Validate the given origin for this site.
+
+        Check than the origin looks valid and matches the origin pattern in
+        specified list `` allowed_origins``. Any pattern begins with a scheme.
+        After the scheme there must be a domain. Any domain beginning with a period
+        corresponds to the domain and all its subdomains (for example, ``http://.example.com``
+        ``http://example.com`` and any subdomain).After the domain there must be a port,
+        but it can be omitted. `` * `` matches anything and anything
```
andrewgodwin: No spaces around the `*` here I think.

channels/security/websocket.py
```diff
@@ -21,27 +24,134 @@ def __call__(self, scope):
...
+
+    def validate_origin(self, parsed_origin):
+        """
+        Validate the given origin for this site.
+
+        Check than the origin looks valid and matches the origin pattern in
+        specified list `` allowed_origins``. Any pattern begins with a scheme.
+        After the scheme there must be a domain. Any domain beginning with a period
+        corresponds to the domain and all its subdomains (for example, ``http://.example.com``
+        ``http://example.com`` and any subdomain).After the domain there must be a port,
```
andrewgodwin: Space after the `.` needed

channels/security/websocket.py
```diff
@@ -21,27 +24,134 @@ def __call__(self, scope):
...
+        but port is optional.
+
+        Returns `` True`` for a valid host, `` False`` otherwise.
+        """
+        return any(
+            pattern == "*" or self.is_allowed_origin(parsed_origin, pattern)
+            for pattern in self.allowed_origins
+        )
+
+    def is_allowed_origin(self, parsed_origin, pattern):
```
andrewgodwin: I would call this `match_allowed_origin` so it's obviously different from `validate_origin`

channels/security/websocket.py
```diff
@@ -21,27 +24,134 @@ def __call__(self, scope):
...
+            pattern_hostname = urlparse("//" + pattern).hostname or pattern
+            return is_same_domain(parsed_origin.hostname, pattern_hostname)
+        # Get origin.port or default ports for origin or None
+        origin_port = self.get_origin_port(parsed_origin)
+        # Get pattern.port or default ports for pattern or None
+        pattern_port = self.get_origin_port(parsed_pattern)
+        # Compares hostname, scheme, ports of pattern and origin
+        if parsed_pattern.scheme == parsed_origin.scheme and origin_port == pattern_port \
+                and is_same_domain(parsed_origin.hostname, parsed_pattern.hostname):
+            # Check ssl security certificate
```
andrewgodwin: I do not think we should try to check the certificate in this code at all, and definitely not by default.

Chukak:
I updated the code in accordance with your recommendations. Yes, you were right, `` * .example.com`` is a backwards incompatible change, I added tests for this.

docs/topics/security.rst
```diff
@@ -37,10 +39,17 @@ valid domains as the second argument::
...
                 ])
             ),
-            ["goodsite.com", "*.goodsite.com"],
+            [".goodsite.com", "http://.goodsite.com:80", "http://other.site.com"],
         ),
     })
 
+Note: If you want to resolve any domains, just add ``["*"]`` as second argument.
+
+``OriginValidator`` can accept the ``check_cert`` as fourth argument.
```
andrewgodwin: `check_cert` is no longer there

docs/topics/security.rst
```diff
@@ -25,7 +25,9 @@ ASGI middlewares it contains, ``OriginValidator`` and
 ``OriginValidator`` lets you restrict the valid options for the ``Origin``
 header that is sent with every WebSocket to say where it comes from. Just wrap
 it around your WebSocket application code like this, and pass it a list of
-valid domains as the second argument::
+valid domains as the second argument. You can pass only domain (for example,
```
andrewgodwin: "only a single domain"

docs/topics/security.rst
```diff
@@ -25,7 +25,9 @@ ASGI middlewares it contains, ``OriginValidator`` and
...
+``.allowed-domain.com``) and full origin, than must be scheme://domain[:port]
```
andrewgodwin: "and full origin, than must be" -> "or a full origin, in the format" (and make the format code-style formatting)

docs/topics/security.rst
```diff
@@ -37,10 +39,17 @@ valid domains as the second argument::
                     ...
                 ])
             ),
-            ["goodsite.com", "*.goodsite.com"],
+            [".goodsite.com", "http://.goodsite.com:80", "http://other.site.com"],
         ),
     })
 
+Note: If you want to resolve any domains, just add ``["*"]`` as second argument.
```
andrewgodwin: "any domains, just add * as second argument" -> "any domain, then use the origin `*`"

Chukak:
I updated the docs for websockets.

andrewgodwin:
Thanks so much!

PR #945, opened 2018-03-02
Added imports for Websocket consumers

Bayoslav:
The example code should have an import for 'new' things it imports, makes it much easier for new programmers.


andrewgodwin:
Agreed - I have been trying to add these too as I notice they're missing. Thanks!

Issue #870, opened 2018-02-06
Allow URLRouter to nest inside itself


andrewgodwin:
Right now, URLRouter does not work "as expected" if it's nested inside itself for people used to Django's URL routing and `include()` - it matches the whole URL every time, at every level.

Instead, it should only match on unmatched portions from previous URLRouters (if one is higher in the stack than it), and should combine captured positional/named groups. 

There's an xfailed test in place with desired behaviour: https://github.com/django/channels/blob/08047dddf65f44bf4312e639ad0009bd1ab6f837/tests/test_routing.py#L88

I'd recommend transporting the remaining path to be matched in the scope as a separate entry called `remaining_path`, having the router look for this before using `scope["path"]`, and then writing the rest of its URL out to that scope entry. Once that's done, unmark the test for xfail, add any extra tests that are needed, and update the routing docs to mention the behaviour.

chuckoy:
Hi Andrew,

I've tried running the failing test locally and listing the routes but I can't seem to access the `outer_router` route from `inner_router`.

How should I go about making calls to parent routes on partial matches?

Am I correct in that it ought to try matching the path left to right, while taking away positional/named groups as matches are made?

andrewgodwin:
You shouldn't need to access the outer router from the inner - the outer router should match as much as it can, and then put the remainder of the unmatched path into the `scope` for the inner router to look at. If you take a look at the URL code in Django itself you'll see it has a way of working out what part of the URL is left "unmatched" (as in, where the regular expression/path stops)

hishnash:
Is it not possible to make the `URLRouter` just support nested url routes rather than nesting `URLRouter`s

andrewgodwin:
Maybe, but people are going to nest `URLRouter` anyway because you can (and because it's a valid ASGI app) so I'd rather fix the core issue.

hundredeir:
I dunno why didn't they return the new path.[link](https://github.com/django/django/blob/d4373b6da4b420fe9211438addeedb396a3821be/django/urls/resolvers.py#L342)

andrewgodwin:
Right, that's probably one of the changes to make this work.

matthiask:
I'm a bit surprised that the desired behaviour includes capturing both named and unnamed groups. The Django docs say:

> In either case, using only one style within a given regex is recommended. When both styles are mixed, any unnamed groups are ignored and only named groups are passed to the view function.

[...]

PR #946, opened 2018-03-03
Add support for nested URLRouter instances

matthiask:
Refs #870

Instead of modifying `scope["path"]` I introduced a new scope variable `"path_remaining"` for inner URLRouter instances, which -- if present -- is preferred by URLRouter to `"path"`.

The `test_url_router_nesting` test is modified to not expect positional AND keyword arguments, because Django also ignores all positional arguments as soon as there are any keyword arguments in URLconfs.

### DIFF
--- a/channels/routing.py
+++ b/channels/routing.py
@@ -75,14 +75,21 @@ class URLRouter:
         # Run through the routes we have until one matches
         for route in self.routes:
             try:
-                match = route.resolve(path)
-                if match is not None:
+                match = route.pattern.match(path)
+                if match:
+                    new_path, args, kwargs = match
+                    # Shallow copy of scope.
+                    scope = dict(scope)
+                    outer = scope.get("url_route", {})
+                    # Only pass on the rest of the path
+                    scope["path"] = new_path
+                    # scope["path"] = scope["path"][match.end():]
                     # Add args or kwargs into the scope
                     scope["url_route"] = {
-                        "args": match.args,
-                        "kwargs": match.kwargs,
+                        "args": outer.get("args", ()) + args,
+                        "kwargs": {**outer.get("kwargs", {}), **kwargs},
                     }
-                    return match.func(scope)
+                    return route.callback(scope)
             except Resolver404 as e:
                 pass
         else:
--- a/channels/routing.py
+++ b/channels/routing.py
@@ -53,6 +53,25 @@ class ProtocolTypeRouter:
             raise ValueError("No application configured for scope type %r" % scope["type"])
 
 
+def route_pattern_match(route, path):
+    """
+    Backport of RegexPattern.match for Django versions before 2.0. Returns
+    the remaining path and positional and keyword arguments matched.
+    """
+    if hasattr(route, "pattern"):
+        return route.pattern.match(path)
+    # Django<2.0. No converters... :-(
+    match = route.regex.search(path)
+    if match:
+        # If there are any named groups, use those as kwargs, ignoring
+        # non-named groups. Otherwise, pass all non-named arguments as
+        # positional arguments.
+        kwargs = match.groupdict()
+        args = () if kwargs else match.groups()
+        return path[match.end():], args, kwargs
+    return None
+
+
 class URLRouter:
     """
     Routes to different applications/consumers based on the URL path.
@@ -75,7 +94,7 @@ class URLRouter:
         # Run through the routes we have until one matches
         for route in self.routes:
             try:
-                match = route.pattern.match(path)
+                match = route_pattern_match(route, path)
                 if match:
                     new_path, args, kwargs = match
                     # Shallow copy of scope.
--- a/channels/routing.py
+++ b/channels/routing.py
@@ -99,11 +99,10 @@ class URLRouter:
                     new_path, args, kwargs = match
                     # Shallow copy of scope.
                     scope = dict(scope)
-                    outer = scope.get("url_route", {})
                     # Only pass on the rest of the path
                     scope["path_remaining"] = new_path
-                    # scope["path"] = scope["path"][match.end():]
                     # Add args or kwargs into the scope
+                    outer = scope.get("url_route", {})
                     scope["url_route"] = {
                         "args": outer.get("args", ()) + args,
                         "kwargs": {**outer.get("kwargs", {}), **kwargs},
--- a/channels/routing.py
+++ b/channels/routing.py
@@ -67,7 +67,7 @@ class URLRouter:
 
     def __call__(self, scope):
         # Get the path
-        path = scope.get("path", None)
+        path = scope.get("path_remaining", scope.get("path", None))
         if path is None:
             raise ValueError("No 'path' key in connection scope, cannot route URLs")
         # Remove leading / to match Django's handling
@@ -82,7 +82,7 @@ class URLRouter:
                     scope = dict(scope)
                     outer = scope.get("url_route", {})
                     # Only pass on the rest of the path
-                    scope["path"] = new_path
+                    scope["path_remaining"] = new_path
                     # scope["path"] = scope["path"][match.end():]
                     # Add args or kwargs into the scope
                     scope["url_route"] = {
--- a/channels/routing.py
+++ b/channels/routing.py
@@ -81,8 +81,14 @@ class URLRouter:
     url() or path().
     """
 
+    extensions = {"path_routing"}
+
     def __init__(self, routes):
         self.routes = routes
+        if self.routes and hasattr(self.routes[0], "pattern"):
+            for route in self.routes:
+                if "path_routing" in getattr(route.callback, "extensions", set()):
+                    route.pattern._is_endpoint = False
 
     def __call__(self, scope):
         # Get the path
--- a/channels/routing.py
+++ b/channels/routing.py
@@ -81,13 +81,19 @@ class URLRouter:
     url() or path().
     """
 
-    extensions = {"path_routing"}
+    #: This router wants to do routing based on scope[path] or
+    #: scope[path_remaining]. ``path()`` entries in URLRouter should not be
+    #: treated as endpoints (ended with ``$``), but similar to ``include()``.
+    _path_routing = True
 
     def __init__(self, routes):
         self.routes = routes
+        # Django 2 introduced path(); older routes have no "pattern" attribute
         if self.routes and hasattr(self.routes[0], "pattern"):
             for route in self.routes:
-                if "path_routing" in getattr(route.callback, "extensions", set()):
+                # The inner ASGI app wants to do additional routing, route
+                # must not be an endpoint
+                if getattr(route.callback, "_path_routing", False) is True:
                     route.pattern._is_endpoint = False
 
     def __call__(self, scope):
--- a/docs/topics/routing.rst
+++ b/docs/topics/routing.rst
@@ -103,13 +103,18 @@ a single argument, a list of Django URL objects (either ``path()`` or ``url()``)
     ])
 
 Any captured groups will be provided in ``scope`` as the key ``url_route``, a
-dict with an ``args`` key containing a list of positional regex groups and a
-``kwargs`` key with a dict of the named regex groups.
+dict with a ``kwargs`` key containing a dict of the named regex groups and
+an ``args`` key with a list of positional regex groups. Note that named
+and unnamed groups cannot be mixed: Positional groups are discarded as soon
+as a single named group is matched.
 
 For example, to pull out the named group ``stream`` in the example above, you
 would do this::
 
     stream = self.scope["url_route"]["kwargs"]["stream"]
+
+Please note that ``URLRouter`` nesting will not work properly with
+``path()`` routes if inner routers are wrapped by additional middleware.
 
 
 ChannelNameRouter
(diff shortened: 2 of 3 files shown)


matthiask:
I added additional tests. Unfortunately, `path` cannot be used for the outer URLRouter easily because the ASGI app is callable, and therefore Django thinks that it is a view and appends a `$` to the regex.

andrewgodwin:
Ooh, that `path` thing is annoying, I forgot that was how it distinguished between views and includes. Ideally I'd like to make it work, though I accept that might be messy.

Is there any way you see of making it work? Or would it be too hard to reach inside and undo the `is_endpoint=True` damage to the regex?

matthiask:
[Here](https://github.com/matthiask/channels/tree/nested-urlrouter-path)'s a branch with an additional commit on top which makes the tests pass using `path()` in the outer URLRouter:

https://github.com/matthiask/channels/commit/cdd335be2e7c5a1d1049cc5b6f2b8a976cff07e9#diff-1a24e8059b03907de2c85bc441155308R85

It feels like a really bad idea though. The outer URLRouter would have to know that somewhere inside there's another ASGI app which wants to do some routing with the remaining path. The code in the commit above works only when 1. the inner ASGI app is an URLRouter and 2. there is no intermediate turtle in the stack of turtles. Introducing a way so that outer turtles know about inner turtles sounds very unasgilike (if that was a word).

Here's an alternative idea: Adding a `path()` function to `channels/routing.py` with support for overriding `is_endpoint` (and `re_path()` for consistency) -- no docs, and does not work with Django 1.11 yet:

https://github.com/matthiask/channels/commit/6d9830c9ec9e663da8bc38b45e715e1fcf3f882d

This also feels a bit bad. The `_path` function uses too many Django internals for my taste and I fear breakage down the road. It might be better to contribute this change to Django, but even almost 10 years after the initial contribution I find the contributing process a bit daunting and also we'd have to wait a long time since 1.11 is an LTS release.

andrewgodwin:
Yeah, both of those are too fragile for me, really - thanks for the investigative work though.

I'm just particularly uneasy about shipping this with `path` being a silent mysterious failure (obviously things are worse now, so we need this in some form, but I want to give it a decent level of thought).

What if, instead of trying to fix it, we just detect when it happens and raise an error? I suspect having an inner URLRouter detect that it is somewhere inside another URLRouter is quite easy (because of the presence of things in the `scope`), which means we just need to make sure that we can detect when there's no URL remaining to match on for the inner one and raise an error. This would also not work if someone deliberately made an inner URLRouter that just matched on `""` but... they should just not do that in that case.

matthiask:
Hmm. I'm not sure if I follow.

`URLRouter.__init__` does not have access to the `scope`, so we cannot signal to inner `URLRouter`s that they should check the outer routers' routes.

`URLRouter.__call__` of the inner router is never run when the outer `URLRouter` uses `path` because of the `$` in the regex, which seems only possible with fragile hacks.

I don't see a way around either changing Django's `path()` function or documenting this as a gotcha. It's a pity. `re_path()` isn't even strictly more powerful because of the converters that `path()` supports.

--- 

Here's a third version. This one modifies the routes passed to `URLRouter` upon initialization. Probably not better, since it uses the same trick as version 1 above, but only runs once.

https://github.com/matthiask/channels/commit/84c5dea7b365cc3733a099fa082758d1c854693f

andrewgodwin:
Oh, right, the inner one is never even called as it won't match. Hrmph.

That third version is not bad, it's only 3 lines of logic and it's not super-fragile. I would be alright with that.

matthiask:
I don't think it's a solution, because the third solution converts *all* routes to open-ended routes -- it apples `is_endpoint=False` everywhere, and does not look whether some inner ASGI app wants to do routing or not.

I'm still thinking about this issue, but haven't found a good solution yet. It's always too early (information isn't here yet) or too late (e.g. signaling to outer URLRouters that no inner ASGI app did any routing via `self.scope` -- but at this time, the event has already been processed in a possibly (probably) non-idempotent way)

andrewgodwin:
Let's not rush and try to make it as safe as possible :) People won't read documentation, of that I can assure you.

I'm much happier with a slightly hackish way of detecting errors, too, rather than a hackish fix. That said, I don't have any better ideas right now either...

matthiask:
Here's another idea: Add an attribute to ASGI apps, `does_path_routing`, which if `True` signals to outer URLRouter instances that it should not treat the app as an endpoint.

The change in version 3 above made the `test_url_router_path` fail because the first `path("")` did match everything. The new idea does not have this problem. 

Change added to this PR, but not documented yet. Relevant commit here: https://github.com/django/channels/pull/946/commits/04f82e95e1db897adee74ebd670382c0151f559f

matthiask:
Maybe `app.extensions = {"path_routing"}` or something would be better. (Similar to the channel layer extensions attribute.)

andrewgodwin:
That has the same issue as the earlier attempt to see if it was a URLRouter, though - it won't work if you've got middleware or similar in the way.

I think this is still probably the best approach, but I'm not sure this is the right place to jump to an `extensions` concept - maybe just have an underscored attribute instead?

matthiask:
Yes, it's true that middleware between URLRouter requires additional work, but setting an attribute seems less work (and also more correct) than inheriting URLRouter just for the `isinstance` check. I added a middleware without additional functionality to the [test](https://github.com/django/channels/pull/946/commits/948f855d729aabde96f17d50e1f8f33ec6f57962#diff-06c0af743b62c3e720fe92d0ce2b7f7bR129) for this reason.

Also, converted `extensions` back to `_path_routing`, and amended the test to `is True` instead of checking for truish values so that MagicMock works.

andrewgodwin:
Lovely, I think this will be the best we can get for now. Do you mind also adding a note to the routing documentation that mentions url nesting won't work properly with `path()` and urlrouters-inside-middleware?

matthiask:
I also reworded the kwargs/args discussion a bit (kwargs are mentioned first because they are preferred, and also mention that args and kwargs can't be mixed)

andrewgodwin:
Great - merged! Thanks for your work on this; with this and a couple other big things landed in the last week I'll get round to doing a new 2.1 release soon.

hoefling:
One question - is the copying of the scope necessary here?

https://github.com/django/channels/blob/984e9dc9cece66f7520bc9c1fd0583e9c60b1022/channels/routing.py#L113

I have a custom middleware that modifies the scope and the scope changes are not propagated to the consumers anymore. Example:

```python
class MyMiddleware:

    def __init__(self, scope: Scope, middleware: Middleware) -> None:
        self.scope = scope
        self.middleware = middleware
        self.inner = self.middleware.inner(scope)
        self.receive_orig = None
        self.send_orig = None

    def __call__(self, receive: Receiver, send: Sender) -> Middleware:
        self.receive_orig = receive
        return self.inner(self.receive, send)

    async def receive(self):
        """Overridden receive that clears stored credentials once the token expires."""
        if self.scope['token'] is not valid:
            self.scope.pop('token', None)
        return await self.receive_orig()

class MyConsumer(AsyncWebsocketConsumer):
    async def receive(self, text_data=None, bytes_data=None):
        if 'token' not in self.scope:
            await self.close()
        else:
            await self.send(text_data='ping')

Removing `token` has no effect in `MyConsumer`. Not sure if this also affects `SessionMiddleware`, but possible.

**Comment by matthiask** [comment]
I did give this particular behavior a lot of thought, but probably not enough. `scope` isn't copied anywhere else, so if possible it shouldn't be copied here either.

I thought that path_remaining should not be visible to other routers and middleware, and the easiest way to ensure this was to make a shallow copy of the scope. 

The current tests all pass without the shallow copy, but I'd like to add some additional tests to annoy the code some more. I should be able to submit another pull request until tomorrow evening (UTC)

**Comment by hoefling** [comment]
@matthiask let's wait for what @andrewgodwin has to say :-) First you clearly had your reasons to implement things as they are, second I'm not even sure the middleware should be allowed to modify the scope - it's just something I assumed based on Django's middleware hooking patterns. If the scope shouldn't be shared on the way in the whole middleware chain, it's also fine, I'll just move the middleware's logic to the consumer itself, maybe introduce some consumer mixin for the reusable code.

**Comment by matthiask** [comment]
It's almost too easy to construct a failing test with the current code:

```
# @pytest.mark.xfail
def test_path_remaining():
    inner_router = URLRouter([
        url(r'^no-match/$', MagicMock(return_value=1)),
    ])
    test_app = MagicMock(return_value=2)
    outer_router = URLRouter([
        url(r'^prefix/', inner_router),
        url(r'^prefix/stuff/$', test_app),
    ])

    assert outer_router({"type": "http", "path": "/prefix/stuff/"}) == 2
```

The two changes that are necessary for this test to pass are the following

- Restore `path_remaining` after calling `route.callback(scope)` 
- Raise a `Resolver404` instead of a `ValueError` if no route of the current router matches, since this exception is handled by an outer URLRouter.

Unfortunately the `Resolver404` exception is not known to Daphne, otherwise this might also help with solving https://github.com/django/daphne/issues/165 (Return 404 for invalid routes)

The WIP code is available here: https://github.com/matthiask/channels/commit/bd9bda78a8ad876493357b923012ec9feb60f6a3

**Comment by andrewgodwin** [comment]
I had always in my head assumed that scope was something that was immutable once it was handed down through the ASGI first-call, so I was for adding `copy` in though. You are right, however, that we should at least be consistent about it - either make all other middlewares we ship also copy, or remove the copy here.

The reason I would probably vote for everything copying is that modifying scope in a middleware like in the example above feels wrong - it's not meant to be something you check dynamically, but more like a Django HttpRequest where you can tack on attributes but you're not meant to change it in a major way. Thoughts?

**Comment by matthiask** [comment]
I think copying makes more sense, yes. The `scope` comes from outside; if a middleware or a router wants to save additional data somewhere, it can just add attributes to `self`.

I force updated the branch referenced above; to me the code looks much better like this:
https://github.com/django/channels/compare/master...matthiask:more-urlrouter-nesting

Should I prepare a pull request implementing the scope copying change? It would involve changing a few places in the documentation too:

```
~/Projects/channels/channels$ git grep -E 'scope\[.* = '
channels/auth.py:    scope["user"] = user
channels/auth.py:        scope["user"] = AnonymousUser()
channels/auth.py:            scope["user"] = SimpleLazyObject(lambda: async_to_sync(get_user)(scope))
channels/http.py:            self.scope["headers"] = [
channels/sessions.py:                scope["cookies"] = parse_cookie(value.decode("ascii"))
channels/sessions.py:            scope["cookies"] = {}
channels/sessions.py:        self.scope["session"] = self.middleware.session_store(session_key)
channels/staticfiles.py:            scope["static_base_url"] = self.base_url
docs/topics/authentication.rst:            scope["user"] = User.objects.get(id=int(scope["query_string"]))
docs/topics/sessions.rst:            self.scope["session"]["seed"] = random.randint(1, 1000)
~/Projects/channels/channels$ 
```

(Idle thought: In some ways `scope` is a bit misleading to me, because it sounds like it would be the preferred place to put variables. It is more similar to an `environ`ment for inner ASGI apps. Modifying a scope is what we do all the time when programming, modifying the environment is a thing which requires more thought.)

**Comment by hoefling** [comment]
Nice to see a decision here! I would also request some explaining note regarding it in the docs as it seems I'm not the only one being misled, see the PR review [here](https://github.com/django/channels/pull/973/files#r174378332).

**Comment by ovekaaven** [comment]
This seems to introduce a regression for me:

```
2018-04-01 00:44:45,441 - ERROR - ws_protocol - Traceback (most recent call last):
  File "/home/ovek/django-xmpp-server/example/env/lib/python3.6/site-packages/daphne/ws_protocol.py", line 76, in onConnect
    "subprotocols": subprotocols,
  File "/home/ovek/django-xmpp-server/example/env/lib/python3.6/site-packages/daphne/server.py", line 184, in create_application
    application_instance = self.application(scope=scope)
  File "/home/ovek/django-xmpp-server/example/chatbox/channels/staticfiles.py", line 41, in __call__
    return self.application(scope)
  File "/home/ovek/django-xmpp-server/example/chatbox/channels/routing.py", line 51, in __call__
    return self.application_mapping[scope["type"]](scope)
  File "/home/ovek/django-xmpp-server/example/chatbox/channels/routing.py", line 119, in __call__
    "kwargs": {**outer.get("kwargs", {}), **kwargs},
TypeError: 'NoneType' object is not callable
```

I'm using includes in urls:

```
ws_urls = [
    url(r'^bind/$', WSConsumer, name='ws'),
]

    'http': URLRouter([
        url(r'', AsgiHandler)
    ]),
    'websocket': URLRouter([
        url(r'^chat/', include(ws_urls))
    ])
})

```

matthiask:
@ovekaaven Indeed it does. There is no test in channels for `include()` inside `URLRouter()`, so this has not been detected earlier. 

Failing test here: https://github.com/matthiask/channels/commit/accf656a717e606f4043da26181b9a2ff1acf565  I'll see whether I can fix this.

andrewgodwin:
I did not intend for includes to work, which is why there is no test for them.

ovekaaven:
If you don't want them to work, then you should at least update the documentation.

andrewgodwin:
I'm not saying I don't want them to, just that I never got round to documenting and testing either way. It looks like support may be possible given #996, it's now a question of if it makes sense to offer two ways of nesting.

PR #947, opened 2018-03-03
Remove the unused testing_settings.py file

matthiask:


andrewgodwin:
Thanks for spotting that.

PR #948, opened 2018-03-03
Make it even easier to locally run style checks, compile docs and run tests

matthiask:
Sorry for the amount of pull requests and related notifications. But anyway, here's a thing I'm using in a few projects of my own. This would have made adapting to " instead of ' easier for me (and maybe for others as well): Using `tox` to run style checks, docs and tests locally.

(Some projects use tox for everything including a local build matrix, but I think this is overkill, I'd rather rely directly on `.travis.yml` there instead of keeping `tox.ini` in sync as well.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,8 @@ __pycache__/
 .tox/
 *.swp
 *.pyc
-.coverage.*
+.coverage*
+.pytest_cache
 TODO
 node_modules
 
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Hmm, I actually just have these in a pre-commit hook locally, ideally I'd like something a bit simpler than having to install tox (I removed it for testing).

matthiask:
Oh, that works too. I replaced the code with a paragraph in `CONTRIBUTING.rst`

andrewgodwin:
Thanks, that'll work for now. Ideally I'd like to include a prebuilt hook at some point, but at least this way people can lint locally if they want.

PR #950, opened 2018-03-04
Remove all remaining uses of six

matthiask:


andrewgodwin:
Oh, damn, I did this already after I saw it was in there but before I saw this PR. Sorry!

PR #951, opened 2018-03-05
 Make clear how to send an event from a synchronous environment

facucosta:
It probably is obvious to people that are familiar with Django Channels 2 that, since background tasks work with channel_layers you have to send events in the same manner as broadcast messages (using the async_to_sync wrapper from a synchronous environment) but for a first-time developer it might be useful to state it in this section.


andrewgodwin:
Good point - best to repeat these things where they're needed. Thanks!

PR #968, opened 2018-03-09
Use raw strings as url() routes

matthiask:
Does not fix any bugs (right now), but follows Django's documentation which also seems to prefer raw strings as arguments to `url()` resp. `re_path()`


andrewgodwin:
Yup, agreed. Thanks.

PR #971, opened 2018-03-12
[FIX] channel layer's one tutorial command is wrong.

EJSohn:
python version 3.5.x
django version 2.0.3
channels version 2.0.2

```
async_to_sync(channel_layer.send)('test_channel', {type: 'hello'})
``` 
command keeps throwing  `TypeError: can't serialize <class 'type'>`. 

```
async_to_sync(channel_layer.send)('test_channel', {'type': 'hello'})
``` 
works well.


andrewgodwin:
Thanks for catching that.

PR #974, opened 2018-03-15
Add `receive_nothing` to the docs

dgilge:
Documentation for PR https://github.com/django/asgiref/pull/37

### DIFF
--- a/docs/topics/testing.rst
+++ b/docs/topics/testing.rst
@@ -50,7 +50,11 @@ ApplicationCommunicator
 -----------------------
 
 ``ApplicationCommunicator`` is the generic test helper for any ASGI application.
-It gives you two basic methods - one to send events and one to receive events.
+It provides several basic methods for interaction as explained below.
+
+You should only need this generic class for non-HTTP/WebSocket tests, though
+you might need to fall back to it if you are testing things like HTTP chunked
+responses or long-polling, which aren't supported in ``HttpCommunicator`` yet.
 
 .. note::
     ``ApplicationCommunicator`` is actually provided by the base ``asgiref``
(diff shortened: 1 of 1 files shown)


docs/topics/testing.rst
```diff
@@ -61,21 +65,59 @@ To construct it, pass it an application and a scope::
...
+    assert event.get("more_body") is None
+    # Check that there isn't another event
+    assert await communicator.receive_nothing() is True
+    # You could continue to send and receive events
+    # await communicator.send_input(...)
+
+The method has two optional parameters:
+
+* ``timeout`` is ``0.1`` by default to keep tests short when there is no event
+  to receive and should be more than enough time.
```
andrewgodwin: I'd rephrase this a bit. Something like `timeout: number of seconds to wait to ensure the queue is empty. Defaults to 0.1.`

docs/topics/testing.rst
```diff
@@ -61,21 +65,59 @@ To construct it, pass it an application and a scope::
...
+* ``interval`` is the time between two checks for new events. (It will return after
+  a check if there is an event.) The default of ``0.01`` should work well. However,
+  if you want to customize it you might use a low value (``0.001``) if you expect an
+  event and a high value (equal to ``timeout``) if you don't.
```
andrewgodwin: Mention interval is in seconds as well, and drop commentary about how well the default works - just say the default is `0.01` and let people who need to tweak it do so. We don't want to have most people worrying about this value.

dgilge:
I shortened the parameter explanations.

andrewgodwin:
Thanks!

PR #975, opened 2018-03-15
Create shallow copies of "scope" instead of modifying it in place when calling the next ASGI app

matthiask:
Maybe it makes more sense to continue the discussion here instead of on the closed pull request for URLRouter nesting (https://github.com/django/channels/pull/946)

I know that the decision whether scope should be copied or not hasn't been made yet. As mentioned [here](https://github.com/django/channels/pull/946#issuecomment-372929152) I think that copying is the right thing to do. 

This pull request includes

- a test for nested URLRouter instances which fails with current master
- scope copying in middleware and routers, no scope copying in utility functions such as `login` and `logout` (since scope is passed as an argument, and supposed to be modified in-place)

### DIFF
--- a/channels/auth.py
+++ b/channels/auth.py
@@ -137,6 +137,7 @@ class AuthMiddleware:
         if "session" not in scope:
             raise ValueError("AuthMiddleware cannot find session in scope. SessionMiddleware must be above it.")
         # Add it to the scope if it's not there already
+        scope = dict(scope)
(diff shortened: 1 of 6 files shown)


andrewgodwin:
Great, thanks!

matthiask:
Thank **you**! It's a pleasure working on this project.

PR #980, opened 2018-03-20
Fix undefined variable reference

GVRV:
G'day! 

I was just reading the introduction and I'm not sure if this patch is the correct behaviour, but I wasn't sure where the `text` variable was coming from. Let me know if this is the wrong patch and you need me to make any corrections, or feel free to ignore this patch altogether if I'm mistaken. 

Thanks!


andrewgodwin:
Nope, you are 100% right. Thanks!

Issue #930, opened 2018-02-24
Add subprotocol support to generic consumers


andrewgodwin:
Subprotocol acceptance is now done on the ASGI application/consumer level, so we should allow `self.accept()` on the consumers to take an optional subprotocol name to accept and pass it back in the ASGI accept message as the `subprotocol` key.

There should also be some basic docs on this, and a simple test.

andrewgodwin:
Feel free to! Let me know if you need help or more of a spec.

PR #988, opened 2018-03-28
Add subprotocol support to generic consumers (#930)

magoogli:
@andrewgodwin Let me know if you would like any other changes. I see that if AcceptConnection is thrown during a connection, accept is also called. I have left it being called without a subprotocol argument, unless you think it should behave in some other way?


andrewgodwin:
I think that's fine for now - this looks lovely as it is (the exception mechanism is not super publicised anyway)!

PR #996, opened 2018-04-01
Urlrouter with include

matthiask:
According to https://github.com/django/channels/pull/946#issuecomment-377734667 the URLRouter changes introduced a regression, where `include()` does not work anymore when used within `URLRouter` routes.

I think we can ignore URL pattern namespaces around `include()` since reversing of channels endpoints isn't implemented (and probably isn't a good fit for the ASGI turtles model anyway)


andrewgodwin:
I actually didn't want to support includes, since this is not a full Django URL implementation, but if this works I'll take it. I would like to see an error handler with the if/elif clause though, in case the object matches neither.

matthiask:
I'm not sure whether it will work with all possible uses of `include()` -- but maybe we should just raise `ImproperlyConfigured` if the second argument to `path()` resp. `url()` isn't callable, and document this? 

It would still be possible for someone to provide an exhaustive test suite for `include()` and the corresponding implementation later.

andrewgodwin:
I would be very much in favour of detecting include and raising a nice error message that says "hey you should nest routers instead"

matthiask:
I replaced the code in the pull request with the `ImproperlyConfigured proposal.

PR #1030, opened 2018-04-24
Pass default args from urlconf entries while matching urls

bpeschier:
Nested URL routers took over the role of `router.resolve()` on the
URLPattern with a custom function. This function forgot to include
any default args given in the `path()`/`url()` with `kwargs={}`.


bpeschier:
Ah, forgot the tests, will fetch them when I'm near my laptop again.

andrewgodwin:
Sorry, this one somehow got stuck in a spam filter and I never saw it. Thanks!

PR #1043, opened 2018-05-04
correct mistake to make the sentence sensible

ahmedakef:
change first tab to second tab to make the sentence sensible
i think it was written 'first' by mistake


matthiask:
No, it's correct as it is. The sentence says to open a *second* tab, type a message, but the message does not appear in the *first* tab. Changing the latter to "second" as well misses the point of the sentence.

andrewgodwin:
Yup, this sentence is correct, this PR would actually make it incorrect, so closing it.

PR #1114, opened 2018-08-01
bugfix: fix request.path is not combine with script_name

### LINKED ISSUE #1113: scope['path'] should take another name
- Detailed description of the overall behaviour

I am trying to deploy a channel application on a cloud PaaS platform.Howerver, when I use django function to get the absolute URI path, something go wrong.

In my case
Platform hostname: www.paas.com
My application root: www.paas.com/s/app/

```
def home(request):
      logger.info('root path: %s' % request.build_absolute_uri())
```

I found the logger get output 
```
root path: www.paas.com/
```

however, It is excepted to be 
```
root path: www.paas.com/s/app/`
```

- Reasoning why it should be in Channels rather than a third-party app

I check the django code about request path info. It has two attribute: `path` and `path_info`.`path` should be the full path of the URI.In the case , it should be `/s/app/`

- Examples of usage, if possible (it's easier to discuss concrete code examples)

Maybe scope['path'] should change a name?Something like scope['path_info']?

jacky15:
resolve #1113

### DIFF
--- a/channels/http.py
+++ b/channels/http.py
@@ -41,15 +41,15 @@ class AsgiRequest(http.HttpRequest):
             self.path_info = scope['path'][len(self.script_name):]
         else:
             self.path_info = scope['path']
-        
+
         # Path info
         if settings.FORCE_SCRIPT_NAME:
             # django path is different from asgi scope path args, it should combine with script name
             self.path = '%s/%s' % (settings.FORCE_SCRIPT_NAME.rstrip('/'), self.path_info.replace('/', '', 1))
-            
+
         else:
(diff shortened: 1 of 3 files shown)


jacky15:
I have updated my code for the formatting check and add the test case about this pull request.But it seems that there is something wrong with other the test cases?

andrewgodwin:
OK, I've fixed the test failures on master. If you rebase your pull request your tests should also pass.

andrewgodwin:
Thanks! Merged it in, will be out in the next release.

PR #1120, opened 2018-08-11
Allow None if '*' is in allowed_origins. Fixes #1115

### LINKED ISSUE #1115: AuthMiddleware not called during tests
Hello channels devs team,

I am writing the tests for the websocket consumers of my application and I cannot successfully test my authentication process. I use a custom  AuthMiddleware to use tokens instead of cookies ([this one](https://gist.github.com/rluts/22e05ed8f53f97bdd02eafdf38f3d60a)) and it seems that the middleware is not called at all during tests, whereas it is actually called with real-world testing (*i.e.* when running the app).

I cannot say if my problem is related to #903, as I process the login by another way with Django REST Framework. The authentication only works if the user is already connected.

Here is the test:
```python
import pytest
from channels.testing import WebsocketCommunicator
from rest_framework.authtoken.models import Token
from my_app.routing import application

@pytest.mark.asyncio                                                           
@pytest.mark.django_db(transaction=True)                                       
async def test_authentication():
    # create a token                                                           
    user = get_an_user_somehow()
    token = Token.objects.create(user=user)                         
    headers = [(b'authorization', "Token {}".format(token.key).encode())]      
                                                                               
    # create communicator and connect
    communicator = WebsocketCommunicator(application, "/ws/my/path/",  
                                         headers=headers)                      
    connected, _ = await communicator.connect()                                
                                                                               
    assert connected  # fails here
```

According to the AuthMiddleware code, a user will always be added to the scope, even if the authentication fails (an anonymous user is added in this case). But there is no user in my scope after awaiting connect, this is why I believe that the AuthMIddleware is not called by the WebsocketCommunicatior. This does not seem to be a problem from the custom AuthMiddleware side.

What do you think?

* **OS and runtime environment** Linux kernel 4.17.11-1-default, Python 3.6.5;
* **Versions** Channels 2.1.2, Daphne 2.2.0, Django 1.11.13, Django REST Framework 3.6.4, Twisted none, ASGI backend in memory;
* **Running** in dev mode with runserver or with pytest.

pvanagtmaal:
Allow None as parsed_origin only if '*' is in allowed_origins. This does require a None check in `match_allowed_origin()`.

### DIFF
--- a/channels/security/websocket.py
+++ b/channels/security/websocket.py
@@ -45,8 +45,8 @@ class OriginValidator:
 
         Returns ``True`` if validation function was successful, ``False`` otherwise.
         """
-        # None is not allowed
-        if parsed_origin is None:
+        # None is not allowed unless all hosts are allowed
+        if parsed_origin is None and "*" not in self.allowed_origins:
             return False
         return self.validate_origin(parsed_origin)
 
@@ -85,9 +85,12 @@ class OriginValidator:
         ``http://exapmple.com``). After the domain there must be a port,
         but it can be omitted.
 
-        Note. This function assumes that the given origin has a schema, domain and port,
-        or only domain. Port is optional.
+        Note. This function assumes that the given origin is None, has a schema,
+        domain and port, or only domain. Port is optional.
         """
+        if parsed_origin is None:
+            return False
+
         # Get ResultParse object
         parsed_pattern = urlparse(pattern.lower(), scheme=None)
         if parsed_pattern.scheme is None:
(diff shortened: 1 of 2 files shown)


channels/security/websocket.py
```diff
@@ -85,9 +85,12 @@ def match_allowed_origin(self, parsed_origin, pattern):
         ``http://exapmple.com``). After the domain there must be a port,
         but it can be omitted.
 
-        Note. This function assumes that the given origin has a schema, domain and port,
-        or only domain. Port is optional.
+        Note. This function assumes that the given origin is None, has a schema,
```
andrewgodwin: This reads confusingly - I suggest `is either None, a schema-domain-port string, or just a domain string`

pvanagtmaal:
@andrewgodwin Agreed, I've put your suggestion there instead.

andrewgodwin:
Great, thanks for this!

PR #1124, opened 2018-08-15
Alternative web servers

tomchristie:
Add `uvicorn` and `hypercorn` to the deployment docs.

We can probably do more with this, but here's a first pass.
I've tried to give a bit of background context too.


andrewgodwin:
Looks good. If anything, I'm up for giving uvicorn more prominent placement and an equivalent section to Daphne further up if/when you want.

PR #1130, opened 2018-08-22
Remove leading white spaces in cookie value 

ostcar:
MorselCoockie.output() adds a leading whitespace at the cookie-value, see https://github.com/python/cpython/blob/3.7/Lib/http/cookies.py#L371

At least [uvicon](https://www.uvicorn.org/) with [h11](https://h11.readthedocs.io/en/latest/) has a problem with that whitespace. So it has to be stripped.


andrewgodwin:
Great, thanks!

PR #1136, opened 2018-08-29
Caching Django middleware on AsgiHandler class

### LINKED ISSUE #1121: Performance drop when using Channels 2 with WhitenoiseMiddleware
I've just upgraded my Django project's Channels from 1.x to 2.x (there seems to be an issue with 1.x that it creates new session for every connection), replacing my wsgi application with an asgi one + all other changes required by docs. 

After that, my response time for pure-html pages dropped from ~20ms to about 400ms. (I've tested on a real simple one, next to no db involved, no javascript evaluated)

I've fiddled with settings and found out that when I disable `whitenoise.middleware.WhiteNoiseMiddleware` , response times are back to normal. (both `whitenoise==4.0` and `whitenoise==3.3.1`)

In Channels 1.x, inclusion / exclusion did not seem to make an impact (not measurable anyways). 

I run the project using simple `python manage.py runserver`, but using `daphne` produces same results.

When I tried to add some logging to the middleware itself, it seems like the `__call__` itself is not performing bad (around 10ms), but the performance drop is roughly dependent on number of static files.

I'm still in process of stripping my project to some bare version that I'm able to send (it's rather big now).

majgaard:
Closes #1121

### DIFF
--- a/channels/http.py
+++ b/channels/http.py
@@ -255,7 +255,10 @@ class AsgiHandler(base.BaseHandler):
         """
         # Because we create an AsgiHandler on every HTTP request
         # we need to preserve the Django middleware chain once we load it.
-        if self.__class__._middleware_chain:
+        if (
+            hasattr(self.__class__, "_middleware_chain")
+            and self.__class__._middleware_chain
+        ):
             self._middleware_chain = self.__class__._middleware_chain
             self._view_middleware = self.__class__._view_middleware
             self._template_response_middleware = (
--- a/channels/http.py
+++ b/channels/http.py
@@ -249,6 +249,29 @@ class AsgiHandler(base.BaseHandler):
                 content_type="text/plain",
             )
 
+    def load_middleware(self):
+        """
+        Loads the Django middleware chain and caches it on the class.
+        """
+        # Because we create an AsgiHandler on every HTTP request
+        # we need to preserve the Django middleware chain once we load it.
+        if self.__class__._middleware_chain:
+            self._middleware_chain = self.__class__._middleware_chain
+            self._view_middleware = self.__class__._view_middleware
+            self._template_response_middleware = (
+                self.__class__._template_response_middleware
+            )
+            self._exception_middleware = self.__class__._exception_middleware
+
+        else:
+            super(AsgiHandler, self).load_middleware()
+            self.__class__._middleware_chain = self._middleware_chain
+            self.__class__._view_middleware = self._view_middleware
+            self.__class__._template_response_middleware = (
+                self._template_response_middleware
+            )
+            self.__class__._exception_middleware = self._exception_middleware
+
     @classmethod
     def encode_response(cls, response):
         """
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -1,6 +1,6 @@
 import re
 import unittest
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 from django.http import HttpResponse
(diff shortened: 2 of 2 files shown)


majgaard:
Oops. Tests are failing for Django<2.0

channels/http.py
```diff
@@ -249,6 +249,29 @@ def handle_uncaught_exception(self, request, resolver, exc_info):
...
+        """
+        # Because we create an AsgiHandler on every HTTP request
+        # we need to preserve the Django middleware chain once we load it.
+        if self.__class__._middleware_chain:
+            self._middleware_chain = self.__class__._middleware_chain
+            self._view_middleware = self.__class__._view_middleware
+            self._template_response_middleware = (
+                self.__class__._template_response_middleware
+            )
+            self._exception_middleware = self.__class__._exception_middleware
```
michael-k: Is this really necessary?

majgaard: I fear it is. 
Doing the following instead:
```python
    def load_middleware(self):
        """
        Loads the Django middleware chain and caches it on the class.
        """
        # Because we create an AsgiHandler on every HTTP request
        # we need to preserve the Django middleware chain once we load it.
        if (
            not hasattr(self.__class__, "_middleware_chain")
            or self.__class__._middleware_chain is None
        ):
            super(AsgiHandler, self).load_middleware()
            self.__class__._middleware_chain = self._middleware_chain
            self.__class__._view_middleware = self._view_middleware
            self.__class__._template_response_middleware = (
                self._template_response_middleware
            )
            self.__class__._exception_middleware = self._exception_middleware
```
yields this error:
```
ERROR    Exception inside application: inner() takes 1 positional argument but 2 were given
   File "/usr/local/lib/python3.6/site-packages/channels/http.py", line 195, in __call__
     await self.handle(body)
   File "/usr/local/lib/python3.6/site-packages/asgiref/sync.py", line 108, in __call__
     return await asyncio.wait_for(future, timeout=None)
   File "/usr/local/lib/python3.6/asyncio/tasks.py", line 339, in wait_for
     return (yield from fut)
   File "/usr/local/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
   File "/usr/local/lib/python3.6/site-packages/asgiref/sync.py", line 123, in thread_handler
     return self.func(*args, **kwargs)
   File "/usr/local/lib/python3.6/site-packages/channels/http.py", line 228, in handle
     response = self.get_response(request)
   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 81, in get_response
     response = self._middleware_chain(request)
   inner() takes 1 positional argument but 2 were given
```

michael-k: Ah, I see, at least `_middleware_chain` is a method.

andrewgodwin:
This looks lovely! Did you confirm that it fixes the issue? It would be nice to have a test that confirms it, if you can get that working.

majgaard:
Dropped my response times from 5s to 87ms 😀 Will put together a test, though.

andrewgodwin:
I have no doubt it works, but would like to stop it happening again by mistake :)

majgaard:
Test is written (and passing... phew!) 🎉
I think we're ready to go 😄

andrewgodwin:
Thanks so much for your work on this! I'll try to get a release out this week.

KarlSayle:
Hi @andrewgodwin , has this been released yet? Just having issues with slow response times and would like to see if this will fix this issue. Thanks.

michael-k:
@KarlSayle No, this has not been released yet.  Click on `b0f8d36` in *andrewgodwin merged commit b0f8d36 into django:master* and the commit will tell you if it's part of any tag (= release).

You can also guess it by comparing the date of the merge (2018-09-04) with the date of the last [release](https://github.com/django/channels/releases) (2018-08-17).

KarlSayle:
Thanks @michael-k - Im getting response times around 5s and was just wondering if it was related to this fix. Thanks again.

KarlSayle:
Just to follow up - looks like the offender was 'django_heroku.settings(locals(), databases=False)' in settings.py left over from when it was running on Heroku. Response times look to be sub 1s now. Apologies for taking up your time, just thought Id post this in case it helps anyone else in the future (i know its a very specific case). I have also pushed the static files to S3 but I think django_heroku was to blame.

PR #1141, opened 2018-09-03
Add How to use Channels 2 with Nginx and Supervisor Section to deploy

ruralscenery:
doc for deploy channels 2 with nginx/supervisor that using daphne --fd Num parameter


andrewgodwin:
Thanks!

PR #1144, opened 2018-09-04
Fix library reference in client jsdoc

martoncsikos:
While debugging why none of my options work, I discovered that the reference for the docs of the underlying `reconnecting-websocket` library is pointing to a different library than the one actually being used.


andrewgodwin:
I know this seems weird, but this has actually prompted me to finally remove the JS code from Channels entirely since it's unmaintained and untested these days.

Thanks for the pull request, but sorry I won't merge it, since I'm about to delete all those files!

PR #1156, opened 2018-09-20
Fix server logging

### LINKED ISSUE #612: runserver should log to "django.channels.server" instead of stderr

Sgiath:
Fixes #612


.gitignore
```diff
@@ -13,9 +13,16 @@ __pycache__/
...
+
 # IDE and Tooling files
 .idea/*
 *~
 
 # macOS
 .DS_Store
+
+# VSCode Tooling files
+.vscode/
```
andrewgodwin: Please add these to your system-local gitignore/gitexclude rather than the project one.

channels/management/commands/runserver.py
```diff
@@ -131,39 +131,34 @@ def log_action(self, protocol, action, details):
...
-                "HTTP %(method)s %(path)s %(status)s [%(time_taken).2f, %(client)s]\n"
-                % details
-            )
+            msg = "HTTP %(method)s %(path)s %(status)s [%(time_taken).2f, %(client)s]"
+
             # Utilize terminal colors, if available
             if 200 <= details["status"] < 300:
                 # Put 2XX first, since it should be the common case
-                msg = self.style.HTTP_SUCCESS(msg)
+                logger.info(msg, details)
```
andrewgodwin: You've lost the colour information - Django runserver still has it last time I checked, so we need to match that.

.gitignore
Sgiath: You mean just `.vscode/` folder or also `Pipfile` and `Pipfile.lock`? I definitely don't want my other `Pipfile` and `Pipfile.lock` files in `.gitignore` :) but I can workaround it if needed.

channels/management/commands/runserver.py
Sgiath: Yeah I missed that. Fixing in next commit

.gitignore
andrewgodwin: Just the vscode one - it's not part of the Python ecosystem directly, and if we had a folder for every editor it would get messy!

Sgiath: Yeah makes sense.
And what about leaving `Pipfile` and `Pipfile.lock` in the repo? It shouldn't affect anyone except for those using Pipenv and for those using it it would make things just easier, don't you think? I can add some section to the Contribution guide how to setup repo using PIpenv or I can move it to another PR.

andrewgodwin: Pipfiles only make sense on end projects, not on libraries, so they don't really belong in this repo.

andrewgodwin:
Thanks!

PR #1158, opened 2018-09-26
Allow consumers to use specific channel layers (#1069)

### LINKED ISSUE #1069: Consumers should be able to listen on non-default channel layers
I have two channel layers defined in settings: "default" and "progress" (with short lifetime of messages)
```
CHANNEL_LAYERS = {
    "default": {
        "BACKEND": "channels_redis.core.RedisChannelLayer",
        "CONFIG": {
            "hosts": [("localhost", 6379)],
        },
    },
    "progress": {
        "BACKEND": "channels_redis.core.RedisChannelLayer",
        "CONFIG": {
            "hosts": [("localhost", 6379)],
            "prefix": "progress",
            "expiry": 5,
        },
    },
}
```

The routing is:
```
...
    "channel": ChannelNameRouter({
        'o-ready': Consumer,
        'o-progress': ProgressConsumer,
    })
})
```

I run two workers: one for "o-ready" channel and one for "o-progress". 
The issue is when the `ProgressConsumer` is instantiated and receiving a message the `self.channel_layer` points to "default" layer instead of "progress" layer. The reason of this the line 42 of channels/consumer.py:

[https://github.com/django/channels/blob/master/channels/consumer.py#L42](url)
```
# Initalize channel layer
self.channel_layer = get_channel_layer()   # <-- line 42
```
It seems the channel_layer should be initialized with the channel_layer from worker to use correct layer when send message from consumer via `self.send`.

Windows, Python 3.6.5, channels 2.1.1, channels-redis 2.1.1, django 2.0.5

ross-weir:
Fixes #1069

### DIFF
--- a/channels/consumer.py
+++ b/channels/consumer.py
@@ -2,6 +2,7 @@ import functools
 
 from asgiref.sync import async_to_sync
 
+from . import DEFAULT_CHANNEL_LAYER
 from .db import database_sync_to_async
 from .exceptions import StopConsumer
 from .layers import get_channel_layer
@@ -30,6 +31,7 @@ class AsyncConsumer:
     """
 
     _sync = False
+    channel_layer_alias = DEFAULT_CHANNEL_LAYER
 
     def __init__(self, scope):
         self.scope = scope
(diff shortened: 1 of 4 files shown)


andrewgodwin:
Looks good, thanks for the speedy turnaround!

Issue #1135, opened 2018-08-28
Cryptic errors accessing self.scope['user'] in Consumers "too early"


wmorrell:
I am upgrading a project from Channels 2.0.2 to 2.1.3, and encountered errors using my websocket consumers that took a while to track down and diagnose. Log output would look something like this:

```
2018-08-28 19:49:48,877 Thread 139644781235080 daphne.ws_protocol ERROR    [Failure instance: Traceback: <class 'NotImplementedError'>: subclasses of LazyObject must provide a _setup() method
/usr/local/lib/python3.6/site-packages/autobahn/websocket/protocol.py:2794:processHandshake
/usr/local/lib/python3.6/site-packages/txaio/tx.py:429:as_future
/usr/local/lib/python3.6/site-packages/twisted/internet/defer.py:151:maybeDeferred
/usr/local/lib/python3.6/site-packages/daphne/ws_protocol.py:77:onConnect
--- <exception caught here> ---
/usr/local/lib/python3.6/site-packages/twisted/internet/defer.py:151:maybeDeferred
/usr/local/lib/python3.6/site-packages/daphne/server.py:186:create_application
/usr/local/lib/python3.6/site-packages/channels/routing.py:56:__call__
/usr/local/lib/python3.6/site-packages/channels/sessions.py:43:__call__
/usr/local/lib/python3.6/site-packages/channels/sessions.py:141:__call__
/usr/local/lib/python3.6/site-packages/channels/sessions.py:163:__init__
/usr/local/lib/python3.6/site-packages/channels/middleware.py:31:__call__
/usr/local/lib/python3.6/site-packages/channels/routing.py:139:__call__
/usr/local/lib/python3.6/site-packages/channels/generic/websocket.py:19:__init__
./edd/notify/consumers.py:49:groups
./edd/notify/backend.py:85:group_names
/usr/local/lib/python3.6/site-packages/django/utils/functional.py:215:inner
/usr/local/lib/python3.6/site-packages/django/utils/functional.py:259:_setup
]
172.18.0.2:60700 - - [28/Aug/2018:19:49:49] "WSDISCONNECT /ws/notify/" - -
```

I found the problem was in accessing any attributes of `self.scope['user']` before the Consumer calls `connect()`. Initially I was seeing this in overriding `__init__` in my Consumer subclass, and debug logging usernames. Then I saw the above traceback, because a property method for `group_names` was using attributes of `self.scope['user']` to generate the iterable of group names, and the base class `__init__` calls this property.

[...]

PR #1161, opened 2018-09-30
throw a more useful error message when scope['user'] is accessed before it's resolved (#1135)

xinbaDev:
Fix proposal for #1135 . Thanks for the review!


andrewgodwin:
Thanks!

PR #1164, opened 2018-10-03
Add 'python_requires' check

OskarPersson:
This should prevent users of Python <3.5 from installing the package


andrewgodwin:
Thanks!

PR #1172, opened 2018-10-19
BugFix: AsyncHttpConsumer

strexicious:
Previously it overrid \_\_call__ causing it to not have channel attributes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ node_modules
 
 # macOS
 .DS_Store
+
+# virtualenv
+env/
--- a/channels/generic/http.py
+++ b/channels/generic/http.py
@@ -1,5 +1,9 @@
 from channels.consumer import AsyncConsumer
 
+from ..exceptions import (
+    StopConsumer,
+)
+
 
 class AsyncHttpConsumer(AsyncConsumer):
     """
@@ -7,25 +11,7 @@ class AsyncHttpConsumer(AsyncConsumer):
     HTTP endpoints.
     """
 
-    async def __call__(self, receive, send):
-        """
-        Async entrypoint - concatenates body fragments and hands off control
-        to ``self.handle`` when the body has been completely received.
-        """
-        self.send = send
-        body = []
-        try:
-            while True:
-                message = await receive()
-                if message["type"] == "http.disconnect":
-                    return
-                else:
-                    if "body" in message:
-                        body.append(message["body"])
-                    if not message.get("more_body"):
-                        await self.handle(b"".join(body))
-        finally:
-            await self.disconnect()
+    body = []
 
     async def send_headers(self, *, status=200, headers=None):
         """
@@ -87,3 +73,22 @@ class AsyncHttpConsumer(AsyncConsumer):
         from here.
         """
         pass
+    
+    async def http_request(self, message):
+        """
+        Async entrypoint - concatenates body fragments and hands off control
+        to ``self.handle`` when the body has been completely received.
+        """
+        try:
+            self.body.append(message["body"])
+            if not message.get("more_body"):
+                await self.handle(b"".join(self.body))
+        finally:
+            await self.disconnect()
+    
+    async def http_disconnect(self, message):
+        """
+        Let the user do their cleanup and close the consumer.
+        """
+        await self.disconnect()
+        raise StopConsumer()


channels/generic/http.py
```diff
@@ -1,31 +1,17 @@
...
-                if message["type"] == "http.disconnect":
-                    return
-                else:
-                    if "body" in message:
-                        body.append(message["body"])
-                    if not message.get("more_body"):
-                        await self.handle(b"".join(body))
-        finally:
-            await self.disconnect()
+    body = []
```
matthiask: This variable will only be initialized once per Python process instead of once per request. 

Suggestion: You could move the initialization inside `async def __call__` (`self.body = []`) and call `super().__call__` so that `AsyncConsumer` can do its thing.

channels/generic/http.py
```diff
@@ -87,3 +73,22 @@ class AsyncHttpConsumer(AsyncConsumer):
...
+        """
+        Async entrypoint - concatenates body fragments and hands off control
+        to ``self.handle`` when the body has been completely received.
+        """
+        try:
+            self.body.append(message["body"])
+            if not message.get("more_body"):
+                await self.handle(b"".join(self.body))
+        finally:
+            await self.disconnect()
```
matthiask: This is too early. You don't want to disconnect after the first body chunk already. Maybe only run `await self.disconnect()` if an exception happens? (Not sure about this.) Also, I suspect you should raise `StopConsumer` here as well, same as in `http_disconnect`

channels/generic/http.py
```diff
@@ -1,31 +1,17 @@
 from channels.consumer import AsyncConsumer
 
+from ..exceptions import (
+    StopConsumer,
+)
```
matthiask: Keep it on one line, hopefully the lint stage will pass then and we'll be able to see the results of the test run on Travis CI :-)

channels/generic/http.py
matthiask: ... to clarify, the original code uses a while loop to receive one or more `http.request` messages. The code you're proposing will only accept one `http.request` message. If the `more_body` flag is set in the first `http.request` message, `handle()` will never be called. That's bad.

strexicious: Yep, I forgot that `finally` would be executed regardless of success. Would this work instead?
```python
        try:
                
                await self.disconnect()
                raise StopConsumer()
        except:
            await self.disconnect()
            raise StopConsumer()
```

matthiask: Yes, I think that's closer to what works.

Two points: 

- The body is optional [according to the spec](https://asgi.readthedocs.io/en/latest/specs/www.html#request), so you should probably leave check in there whether the "body" is set at all.
- Bare `except:` statements are frowned upon. At least use `except Exception` to not catch SystemExit, KeyboardInterrupt etc. (https://docs.python.org/3/library/exceptions.html#exception-hierarchy)

strexicious: Is it sufficient to do the following to check or are there any falsy values that "body" can have?
```python
body = message.get("body")
if body:
    self.body.append(body)
```

strexicious:
@matthiask Thank you for guiding me through. I believe everything should be in order now if you want to review it one last time.

channels/generic/http.py
```diff
@@ -87,3 +69,28 @@ class AsyncHttpConsumer(AsyncConsumer):
...
+        to ``self.handle`` when the body has been completely received.
+        """
+
+        try:
+            if "body" in message:
+                self.body.append(message["body"])
+            if not message.get("more_body"):
+                await self.handle(b"".join(self.body))
+
+                await self.disconnect()
```
seroy:
```suggestion
```

channels/generic/http.py
```diff
@@ -87,3 +69,28 @@ class AsyncHttpConsumer(AsyncConsumer):
...
+                raise StopConsumer()
```
seroy:
```suggestion

```

channels/generic/http.py
```diff
@@ -87,3 +69,28 @@ class AsyncHttpConsumer(AsyncConsumer):
...
+        except Exception:
```
seroy:
```suggestion
        finally:
```

strexicious: @seroy Won't this `finally` be still executed even if there is more body to receive from the request? We only want to disconnect in case there was an error or it has been completed, in the latter case, it could take several events to happen.

seroy: oh, yep! in this case:
```
async def http_request(self, message):
    """
    Async entrypoint - concatenates body fragments and hands off control
    to ``self.handle`` when the body has been completely received.
    """
    if "body" in message:
        try:
        finnaly:
            await self.disconnect()
            raise StopConsumer()
```
Anyway catching Exception almost always is a bad Idea...

strexicious: @seroy Sorry for making you write the code lol, but thanks!

.gitignore
```diff
@@ -19,3 +19,6 @@ node_modules
 
 # macOS
 .DS_Store
+
+# virtualenv
+env/
```
andrewgodwin: Please remove this - it's specific to the way you run the code. I'd advise putting things like this in a global `.gitexclude` file.

PR #1178, opened 2018-10-31
Removing a note which content is false.

madewulf:
Actually, the part 2 of the tutorial makes you do a migration.


andrewgodwin:
Thanks

PR #1182, opened 2018-11-09
Prevent hitting AttributeError when later importing daphne.server past django.setup().

akatrevorjay:
The `daphne` namespace it still kept around, but without `.server`, causing `AttributeError` since an update to `channels`. This alleviates that.

### DIFF
--- a/channels/apps.py
+++ b/channels/apps.py
@@ -5,6 +5,8 @@ from django.apps import AppConfig
 # (e.g. raven does this).
 import daphne.server
 
+assert daphne.server  # pyflakes doesn't support ignores
+
 
 class ChannelsConfig(AppConfig):
 
--- a/channels/apps.py
+++ b/channels/apps.py
@@ -5,8 +5,6 @@ from django.apps import AppConfig
 # (e.g. raven does this).
 import daphne.server
 
-del daphne.server
-
 
 class ChannelsConfig(AppConfig):
 


andrewgodwin:
Can you give a better example of where this fails? I'd ideally like to see a test added that shows what it fixes.

akatrevorjay:
Sure, here's an example. I have an application that spawns a daphne.server.Server from a click command. (Keep in mind, any application that interfaces with daphne.server will run into this problem.)

```python
@cli.command()
@click.option('--port', default=8000, help='API server port.')
@click.option('--host', default='127.0.0.1', help='API server host.')
@click.pass_context
def apiserver(ctx, port, host):
    import django
    django.setup()

    # This is a hacky workaround for hitting AttributeError when importing daphne.server
    import sys
    [sys.modules.pop(k) for k in list(sys.modules) if k.startswith('daphne.') or k == 'daphne']

    import daphne.server

    srv = daphne.server.Server(
        application,
        endpoints=('tcp:port=%s:interface=%s' % (port, host), ),
        # We do *not* want twisted handling our signals for us.
        signal_handlers=False,
        action_logger=None,
        http_timeout=None,
        websocket_timeout=86400,
        websocket_connect_timeout=20,
        ping_interval=20,
        ping_timeout=30,
        root_path="",
        proxy_forwarded_address_header='X-Real-IP',
        proxy_forwarded_port_header='X-Forwarded-Port',
        proxy_forwarded_proto_header='X-Forwarded-Scheme',
        verbosity=1,
        websocket_handshake_timeout=5,
        application_close_timeout=10,
        ready_callable=None,
    )

    try:
        srv.run()
    except KeyboardInterrupt:
        pass
    finally:
        log.info('Done')
```

andrewgodwin:
Ah, right, of course. Thanks!

akatrevorjay:
No prob. Thank you!

Issue #1170, opened 2018-10-16
Add a size limit for request bodies


dav20011:
Upon reading the request body Django's request objects check the content length and compare it to a size limit set in the config:
```
# Limit the maximum request data size that will be handled in-memory.
if (settings.DATA_UPLOAD_MAX_MEMORY_SIZE is not None and
        int(self.META.get('CONTENT_LENGTH') or 0) > settings.DATA_UPLOAD_MAX_MEMORY_SIZE):
    raise RequestDataTooBig('Request body exceeded settings.DATA_UPLOAD_MAX_MEMORY_SIZE.')
```
AsgiRequest objects do not perform this check, which is very dangerous. I was able to spam the server with only a few large request, which made the server incapable of handling any other request.

andrewgodwin:
Definitely agree, this is verging on a DoS attack.

ross-weir:
Hi @andrewgodwin . Could I please have a go at adding this one?

andrewgodwin:
@ross-weir Feel free to have a go and open a pull request with your work!

jpic:
FYI, this is being re-discussed in #1240 which you're invited to review.

PR #1224, opened 2019-01-20
Size limit for request body (#1170)

Zarathustra2:
Hey there, 

currently trying to submit a pr for this issue #1170.

Basically I copied the code from django/http/request.py, which then
throws the RequestDataTooBig. The exception gets caught in the
handle-method of the AsgiHandler. In the end a HttpResponse with a
status code of 413 is returned.

Let me know if I am on the right track.

### DIFF
--- a/channels/http.py
+++ b/channels/http.py
@@ -8,6 +8,7 @@ from io import BytesIO
 from django import http
 from django.conf import settings
 from django.core import signals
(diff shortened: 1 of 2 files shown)


andrewgodwin:
Lovely, that looks good - thanks! (and I see you had a fun run in with the auto linter!)

PR #1235, opened 2019-01-30
Add tox.ini

jpic:
This will help contributors, and for the joy of maintaining software against plenty of versions of other softwares :joy:

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -81,6 +81,18 @@ Then, you can run the tests::
 
     pytest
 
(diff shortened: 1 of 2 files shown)


andrewgodwin:
I actually got rid of `tox` from the repo a while ago, because nobody was using it. If you want to add this could we have some info about how to use it in the README or the docs too?

andrewgodwin:
Thanks. If that one is wrong, hopefully we can push a new one.

PR #1246, opened 2019-02-10
Fixed #1245: Header keys and values must be bytestrings according to the spec

### LINKED ISSUE #1245: Httpconsumer problem
```
class FetchVendorConsumer(AsyncHttpConsumer):
    async def handle(self, body):
        try:
            await self.send_response(200, b"Hello", headers=[
                ("Content-Type", "application/json"),
            ])
        except:
            print('ss')
```
When i send request to this view, it print ss. Not sure how can I get the exception that is being raised? 

I am doing this because when I send request from frontend I don't get any response back.

matthiask:
Closes #1245.

### DIFF
--- a/docs/topics/consumers.rst
+++ b/docs/topics/consumers.rst
@@ -328,7 +328,7 @@ primitives to implement a HTTP endpoint::
         async def handle(self, body):
             await asyncio.sleep(10)
             await self.send_response(200, b"Your response bytes", headers=[
-                ("Content-Type", "text/plain"),
+                (b"Content-Type", b"text/plain"),
             ])
 
 You are expected to implement your own ``handle`` method. The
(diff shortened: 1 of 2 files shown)


matthiask:
By the way, it would of course be possible to add some code to channels itself checking the type of the header keys and values. I'm not convinced it's worth it because Daphne's error message is quite clear.

Also, I'm just fixing my own oversight here. (https://github.com/django/channels/commit/5189b12679fdb14cddf5d69462b3dffb77b296f3)

matthiask:
> > ...it would of course be possible to add some code to channels itself checking the type of the header keys and values...
> 
> Yeah, hmm. I think probably _Out of Scope_ at the moment. I'd rather keep things functional and vanilla right now.

I agree completely.

PR #1272, opened 2019-03-28
allowed setting default message-type handler for consumers

spielmannj:
I have recently stumbled over this problem: it is part of a group that I use to monitor which channels are active. This "admin" channel receives status updates and "admin" type messages. Sending these messages stopped my websocket client without an error message. Here is the offending code:

        async_to_sync(self.channel_layer.group_send)(
            ADMIN_GROUP_NAME,
            {
                'id': self.group_name,
                'type': "admin",
                'msg_type': msg_type,
                'content': content,
            }
        )

After some digging I found out that a consumer that receives a message must handle that message or it will generate a `ValueError`. This ValueError is apparently enough to stop my original consumer.

From the documentation, it was not clear to me that a Consumer must handle all message types it could possibly receive, and raises errors if not.

I have made a change that (a) documents this behaviour and (b) allows developers to override the behaviour by implementing a `default_handler(self, message)` method. This method replicates the current behaviour, so it is backwards compatible. Any message for which no handler was found is passed to this handler. So one could use it to generically handle messages, log unknown message types or simply ignore them.

### DIFF
--- a/channels/consumer.py
+++ b/channels/consumer.py
@@ -72,7 +72,11 @@ class AsyncConsumer:
         if handler:
             await handler(message)
         else:
-            raise ValueError("No handler for message type %s" % message["type"])
+            await self.default_handler(message)
+
+    async def default_handler(self, message):
+        """Overrideable handler for messages of un-implemented type."""
+        raise ValueError("No handler for message type %s" % message["type"])
 
     async def send(self, message):
         """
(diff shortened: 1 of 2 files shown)


spielmannj:
Since this is my first PR, I do not know whether I missed something. There were no tests that covered this part of the code (as far as I could find), so I did not add any. Let me know if there are any other issues with this PR, or whether there is anything I can add or change!

jheld:
Thank you for finding and working through this! Sounds quite helpful!

spielmannj:
Well, it's useful at least to me -- I am using it in my Consumers.

carltongibson (commented):
Hi @spielmannj — Thanks for the input here, but I'm going to say no. I'd rather fail fast, log the error, and make sure I'm handling all the message types the application sends than add this kind of catch-all. (For me it's not a good increase in the API surface area.) I hope that make sense.

PR #1278, opened 2019-04-09
Bump asgiref requirement to 3.x

dblenkus:
Fixes https://github.com/django/channels/issues/1277


dblenkus:
@carltongibson I've updated the PR as discussed in the issue.

PR #1352, opened 2019-09-16
Handle the HTTP request body as a spooled temp file.

hozblok:
Continuation of #1345 and #1251 

Here is an excerpt from the changes:

- code containing improvement - replacement with the spooled temporary file (https://github.com/django/channels/pull/1251#issuecomment-503956365)
- the fix for #1240 .


hozblok:
Hi @carltongibson ,

Your edits are appropriate!
Thanks for your feedback. :+1:

PR #1469, opened 2020-07-15
Compatibility with isort v5.1

smithdc1:
isort v5 seems to be much more stable now, number of issues on their tracker have also reduced significantly from a week ago.


smithdc1:
Fixed #1462 

`black` and `isort` were in conflict so I updated `isort` settings as per the `black` [docs](https://black.readthedocs.io/en/stable/the_black_code_style.html). See the "A compatible `.isort.cfg`" section.

PR #1506, opened 2020-09-30
Update notes on max threads in database docs. 

Zinoex:
Update database documentation to reflect updates in the concurrent.futures.ThreadPoolExecutuor
https://docs.python.org/3/library/concurrent.futures.html


carltongibson (approved):
Thanks @Zinoex!

Issue #1428, opened 2020-03-16
Make nested URLRouter instances work with Django path()s


adamchainz:
PR #946 added support for nested `URLRouter` instances, but without support for `path()` in this case. It would be nice to add that support. There's a lot of discussion on that PR of the limitations to overcome.

PR #1507, opened 2020-10-01
Use path() instead of re_path()

blurrred:
Change re_path() to path() with the appropriate arguments.
Remove the note recommending re_path() 'due to limitation in URLRouter.'

According to https://channels.readthedocs.io/en/latest/topics/routing.html#urlrouter there is nothing that says URLRouter cannot use path() and even recommends using path(). path() is preferable to re_path() due to readability. I have tested these changes and the app works the same.

### DIFF
--- a/docs/tutorial/part_2.rst
+++ b/docs/tutorial/part_2.rst
@@ -233,15 +233,13 @@ Put the following code in ``chat/routing.py``:
 .. code-block:: python
 
     # chat/routing.py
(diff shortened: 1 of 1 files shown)


carltongibson (changes requested):
This is mistake. Nested routers can't use `path()` See #1428. This exact change was made previously but we had to revert it in #1354 because of #1338. 

Thanks though.

PR #1511, opened 2020-10-17
Update JSONAsycnWebSocketConsumer

agriev:
Simple update that provides a user of library with straightforward control on JSON parse errors. and HINT where to look for it.

Could be just mine issue but I've spent some time trying to find out where is my parse exception. By providing async def json_parse_exception_handler(self, e: ValueError): method to AsyncJsonWebsocketConsumer it makes it obvious where to catch the parse exception, which is not tend to stop consumer. just to notify that parsing of JSON failed.

Sorry, I'm new with pull requests and etc. so probably made something wrong

### DIFF
--- a/channels/generic/websocket.py
+++ b/channels/generic/websocket.py
@@ -256,7 +256,10 @@ class AsyncJsonWebsocketConsumer(AsyncWebsocketConsumer):
 
     async def receive(self, text_data=None, bytes_data=None, **kwargs):
         if text_data:
-            await self.receive_json(await self.decode_json(text_data), **kwargs)
+            try:
+                await self.receive_json(await self.decode_json(text_data), **kwargs)
+            except ValueError as e:
+                await self.json_parse_exception_handler(e)
         else:
             raise ValueError("No text section for incoming WebSocket frame!")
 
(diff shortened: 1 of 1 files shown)


agriev:
python -m pytest                                                                     ✔  newDjango3.8.5  1546  23:44:56 20/10/2020
======================================================================================================== test session starts ========================================================================================================
platform darwin -- Python 3.8.5, pytest-6.1.1, py-1.9.0, pluggy-0.13.1
rootdir: /Users/anton/workspace/Django-Channels/channels, configfile: setup.cfg, testpaths: tests
plugins: celery-4.4.7, django-4.0.0, asyncio-0.14.0
collected 61 items                                                                                                                                                                                                                  

tests/test_http.py .                                                                                                                                                                                                          [  1%]
tests/security/test_auth.py ........                                                                                                                                                                                          [ 14%]
tests/test_generic_http.py .                                                                                                                                                                                                  [ 16%]
tests/test_generic_websocket.py ...........                                                                                                                                                                                   [ 34%]
tests/test_http.py ..............                                                                                                                                                                                             [ 57%]
tests/test_inmemorychannel.py ......                                                                                                                                                                                          [ 67%]
tests/test_layers.py ....                                                                                                                                                                                                     [ 73%]
tests/test_routing.py ........                                                                                                                                                                                                [ 86%]
tests/test_testing.py .......                                                                                                                                                                                                 [ 98%]
tests/security/test_websocket.py .                                                                                                                                                                                            [100%]

======================================================================================================== 61 passed in 4.81s =========================================================================================================

I have no idea why travis build failing and local tests passing. any clue?

carltongibson (commented):
Hi @agriev. Thanks for the suggestion. 

Happy for you to add this, but I'm not sure it's worth the documentation, testing and maintenance effort in core. I'd be inclined to just let this raise and throw the error back to the client as the default.

If folks want to subclass and implement receive no problem, but not keen on adding little hooks like this. 

I hope that makes sense. 

(Would need tests and docs if we were to proceed.)

PR #1516, opened 2020-10-19
Remove universal wheels declaration

adamchainz:
Python 2 is no longer supported, so the wheels are no longer universal.


carltongibson (approved):
Yep. Nice!

PR #1521, opened 2020-10-20
Cache dependencies (pip)

smithdc1:
Added cache for pip. As we're using tox, I suspect this will only give a small benefit as the environments will still need to be re-created. 

I added `${{ env.pythonLocation}}-` to the hash. This is intended to get the python version, and so when a small release is issued the cache gets reset.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,6 +25,13 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+      - uses: actions/cache@v2
+        name: Configure pip caching
+        with:
+          path: ~/.cache/pip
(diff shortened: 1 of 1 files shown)


adamchainz:
Thanks for the PR @smithdc1 , sorry it wasn't seen to for... a while. I think we can close it though because:

1. Since this PR was created, `setup-python` added [built-in caching support](https://github.com/actions/setup-python#caching-packages-dependencies), which would be better to use.
2. `tox.ini` is not a complete enough hash key - since it does not specify exact versions, it will mean the cache will include old versions. Using fully specified dependencies for caching would require a bit more work, like I have on my projects with [fully pinned requirements](https://github.com/adamchainz/django-browser-reload/tree/main/requirements).
3. Caching small amounts of pip wheels doesn't speed up tests significantly - caching the virtualenvs in `.tox` would help, but I haven't seen a recipe for that.

If you'd like to try a more advanced pinning + caching PR, please do!

PR #1525, opened 2020-10-20
skip_install on qa

smithdc1:
Ok -- here is a slightly easier change. 

I don't think we need to install the package (and therefore all the dependencies) to run black et al? 

EDIT: 
Pre = [68 seconds](https://github.com/django/channels/runs/1282795278?check_suite_focus=true)
Post = [36 seconds](https://github.com/django/channels/runs/1283181213?check_suite_focus=true)


carltongibson (approved):
Super. 👍

PR #1545, opened 2020-10-28
Update part_1.rst

alejoortiz:
Hello,

I followed the tutorial and realize that there is an error in the documentation regarding HTTP and AsgiHandler. The correct operation, in this case, is that WebSockets will be answered by daphne and HTTP by gunicorn.

I updated tutorial part_1.rst


alejoortiz:
I tested the solution using the following code https://github.com/alejoortiz/tutorial-django-channels.git

carltongibson (commented):
Hi @alejoortiz — these changes are correct for the 3.0 release, which we're preparing currently. 

What version are you using? 2.4.0, which is the latest on PyPI? — in which case you need to make sure you're using the `stable` docs on RTD (or the 2.x version), both of which you can select on the lower left corning on RTD.

alejoortiz:
Hello Carlton, I was using the 2.4.0 version, documentation is fine for 2.X. I will close this pull, thanks a lot for your help.

PR #1565, opened 2020-11-05
Adjusted deployment asgi.py example. 

KStenK:
Docs changes for #1564


carltongibson (approved):
Hey @KStenK. I pushed an edit. Hopefully it makes sense.

PR #1573, opened 2020-11-12
Update testing.rst

XSPGMike:
I was setting up my tests to keep up with channels 3.0 and I incurred in this issue, if I am not getting something about this let me know please.


carltongibson (approved):
Yes, good thanks. 

For a test, likely you can just instantiate the consumer for a single request — like `SimpleWebsocketApp()` — but `as_asgi()` is the right approach. 

Thanks!

PR #1575, opened 2020-11-15
bug in scope refered from self

yuvaraja2303:
bug in scope variable refers in local variable and not in class variable so bug will raise attribute error "scope attribute not in self"


auvipy (approved):

carltongibson (approved):
Yes, thanks.

PR #1600, opened 2020-12-28
Fix typo in consumers documentation

Coronon:
A note in the consumers  documentation contredicted the right phrasing in the database documentation. One does not need to use `database_sync_to_async` in synchronous code.


carltongibson (approved):
Yes, thanks.

PR #1605, opened 2021-01-02
Support nested module attribute in ASGI_APPLICATION setting

mraleson:
Using Django Channels in a project and realized that daphne supports multiple dots in the application path but Channels doesn't.  This change just allows multiple dots in the ASGI_APPLICATION setting path.

### DIFF
--- a/channels/routing.py
+++ b/channels/routing.py
@@ -1,5 +1,6 @@
 import importlib
 import warnings
+from functools import reduce
 
 from asgiref.compatibility import guarantee_single_callable
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
could you please also add test for the proposed changes?

carltongibson (changes requested):
Hi @mraleson — I'm afraid this isn't a complication we'll add here. 

The idea is that this matches Django's `WSGI_APPLICATION`, so you're looking for something like this: 

```
ASGI_APPLICATION = "your_project.asgi.application"
```

Where `your_project.asgi` is the module and `application` is the ASGI callable. 
We don't want to make it more advanced than that.

PR #1606, opened 2021-01-06
Sort release notes by most recent first

adamchainz:


carltongibson (approved):
Yep, ok, super. Thanks.

PR #1677, opened 2021-04-16
Use tox-py in CI

adamchainz:
This is a [tox plugin I just made](https://pypi.org/project/tox-py/) to replace this shell-fu.


carltongibson (approved):
Ok, this looks great.

PR #1684, opened 2021-04-21
Removed async_generator dependency

smithdc1:
I think we can remove this dependency now that only Python 3.6+ is supported?


carltongibson (approved):
Super. 💃

PR #1693, opened 2021-05-05
WIP: docs: add troubleshooting section

yedpodtrzitko:
the `ImproperlyConfigured` exception seems like a quite common error nowadays, so having it mentioned explicitly somewhere in docs might be helpful.

Having the Troubleshooting section won't magically bring attention to it, so mentioning it somewhere explicitly might help too. Tutorial looks like a starting point for everyone, thus referring it right there could be the right place.

Wording might need to be changed in case #1692 will be accepted and merged.


yedpodtrzitko:
moved into #1692

PR #1694, opened 2021-05-06
change 'type' property to match chat_message method

overhead525:
I noticed that the `channel_layer.group_send()` method was sending a message with property:

`
{
    "type": "chat.message"
}
`

In the tutorial, it is mentioned that consumers will match event types with their methods. Since the method was named `chat_message()`, I figured the event type should be the same. In fact, this pattern is shown elsewhere too.

### DIFF
--- a/docs/topics/channel_layers.rst
+++ b/docs/topics/channel_layers.rst
@@ -112,7 +112,7 @@ layer:
     await self.channel_layer.group_send(
         room.group_name,
         {
-            "type": "chat.message",
+            "type": "chat_message",
             "room_id": room_id,
             "username": self.scope["user"].username,
             "message": message,


yedpodtrzitko:
hi,
dots are replaced with underscore automatically so `chat.message` is acceptable to have too. See docs:

https://channels.readthedocs.io/en/stable/topics/channel_layers.html#what-to-send-over-the-channel-layer

> The name of the method will be the `type` of the event with periods replaced by underscores - so, for example, an event coming in over the channel layer with a `type` of `chat.join` will be handled by the method `chat_join`.

overhead525:
Ah okay, my mistake. Thanks for the quick reply, I'll close this PR

PR #1695, opened 2021-05-09
remove unused `extensions` variable

yedpodtrzitko:
I can't find any usage of this. asgiref docs mentions something like [extensions](https://asgi.readthedocs.io/en/latest/extensions.html), but it seems just like a name coincidence.

if this has just a documentation purpose, it could be done as a comment instead?

### DIFF
--- a/channels/layers.py
+++ b/channels/layers.py
@@ -214,8 +214,6 @@ class InMemoryChannelLayer(BaseChannelLayer):
 
     # Channel layer API
 
-    extensions = ["groups", "flush"]
-
     async def send(self, channel, message):
         """
         Send a message onto a (general or specific) channel.


sevdog:
Extensions are documented [here](https://github.com/django/channels/blob/main/docs/channel_layer_spec.rst#extensions) inchannels and [here](https://github.com/django/asgiref/blob/main/specs/asgi.rst#extensions) in asgiref.

It seems that they are not used by channels itself because this list should be checked by applications.

carltongibson (commented):
It's from the spec. https://channels.readthedocs.io/en/latest/channel_layer_spec.html#extensions

Clients can query this to see if the extension is supported

PR #1706, opened 2021-05-26
Fix asgi callable arguments in OriginValidator

fdev:
The single callable in `OriginValidator` uses an argument order that does not match the [ASGI 3 specification](https://asgi.readthedocs.io/en/latest/specs/main.html#applications):

> ASGI applications should be a single async callable:
> ```
> coroutine application(scope, receive, send)
> ```


carltongibson (approved):
Yes, nice. Thanks. 👍

PR #1729, opened 2021-07-11
Document support for Django 3.2

vasilis-vagenas:
Support for Django 3.2 is already included in tox.ini ( https://github.com/django/channels/blob/main/tox.ini )
The same change might probably need to be applied also to release 3.0.4(given: https://github.com/django/channels/blob/3.0.4/tox.ini )


carltongibson (approved):
Yep, great. Thanks.

PR #1734, opened 2021-08-07
docs: instantiate consumers before passing to communicators

MattBroach:
When testing, Channels 3+ (or, more correctly, asgiref 3+) requires consumers to be instantiated (either by calling `as_asgi()` or simply be initializing the class, e.g. `MyConsumer()`) before be passed into communicators.  The testing section in the docs mostly didn't reflect this change in the code examples, so this PR updates all the consumers in the testing examples to call `as_asgi()` before being passed into the communicators.


carltongibson (approved):
Yes, good. Thank you!

PR #1750, opened 2021-09-16
Fix wrong reference to AsyncConsumer in Introduction

DanielSwain:
Change AsyncConsumer to AsyncWebsocketConsumer

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -166,7 +166,7 @@ asynchronous functions, you can write fully asynchronous consumers:
 
 .. code-block:: python
 
-    class PingConsumer(AsyncConsumer):
+    class PingConsumer(AsyncWebsocketConsumer):
         async def websocket_connect(self, message):
             await self.send({
                 "type": "websocket.accept",


carltongibson (commented):
Hi @DanielSwain. 

Is this right? Subclassing `AsyncWebsocketConsumer` you'd implement `receive()` rather than override the `websocket_connect()` and `websocket_receive()` methods, which are implemented for you, and are the benefit of using the generic consumer... 🤔

DanielSwain:
@carltongibson Perhaps.  I only traced the `AsyncConsumer` definition in the code and saw that it didn't have function definitions for either `websocket_connect()` or `websocket_receive()` and so figured that the author of the docs intended `AsyncConsumer` to be `AsyncWebsocketConsumer` since `websocket_connect()` and `websocket_receive()` were specifically overridden in the example provided.  With what you're saying,  however, perhaps that assumption is backwards.  It seems that there is definitely an error in the docs here, so I'll let it to you or the original document writer (if not you) to take it from here.

PR #1774, opened 2021-12-12
Remove the `providing_args` argument from `django.dispatch.Signal` to…

d9pouces:
Remove the `providing_args` argument from `django.dispatch.Signal` to be compatible with Django 4.0.

Quoting the Django doc: _The providing_args argument is deprecated. As it is purely documentational, it has no replacement. If you rely on this argument as documentation, you can move the text to a code comment or docstring._

### DIFF
--- a/channels/signals.py
+++ b/channels/signals.py
@@ -1,7 +1,7 @@
 from django.db import close_old_connections
 from django.dispatch import Signal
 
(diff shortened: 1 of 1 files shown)


carltongibson (approved):
Yep, thanks! 👍

PR #1792, opened 2022-01-22
Reuse invalid_name_error variable

Incred:
Reuse invalid_name_error when raising a TypeError

Fix django#1791


carltongibson (approved):
OK, thanks.

PR #1814, opened 2022-02-23
Doumenttion: Work around testing errors in tutorial

jmacleod:
I am completely new to django and I was following this great channels tutorial. 
https://channels.readthedocs.io/en/stable/tutorial/part_4.html

When I tried to run the test on my mac I kept getting errors.  First they were:
`AttributeError: Can't pickle local object 'convert_exception_to_response.<locals>.inner'`

I believe adding the multiprocessing import and setting it to fork fixed this.  The sites I searched said this was because of how process forking is handled differently on different OSes.  So maybe I should update it to be a comment?

The second error I got was:
`AttributeError: 'WebDriver' object has no attribute 'switch_to_window'`

Maybe the syntax for that call changed since this tutorial was made?

### DIFF
--- a/docs/tutorial/part_4.rst
+++ b/docs/tutorial/part_4.rst
@@ -53,9 +53,11 @@ Put the following code in ``chat/tests.py``:
     from selenium import webdriver
     from selenium.webdriver.common.action_chains import ActionChains
     from selenium.webdriver.support.wait import WebDriverWait
+    import multiprocessing
 
     class ChatTests(ChannelsLiveServerTestCase):
         serve_static = True  # emulate StaticLiveServerTestCase
+        multiprocessing.set_start_method("fork")
 
         @classmethod
         def setUpClass(cls):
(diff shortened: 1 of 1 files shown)


smithdc1:
Thanks for the patch. 

The Python 3.8 docs say.

Changed in version 3.8: On macOS, the spawn start method is now the default. The fork start method should be considered unsafe as it can lead to crashes of the subprocess. See bpo-33725.

https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

I'm therefore not sure we should be advising to over ride the multiprocess start method in this way.

Sorry I don't have a better proposal on what the next steps should be here.

jmacleod:
@smithdc1, thank you for getting back to me.  Maybe there is a better work around and I am just not familiar enough with multiprocessing and channels to know what it is.

Do you have any thoughts about the second part of the PR where  it seems like I needed to switch from:
`self.driver.switch_to_window()`
to 
`self.driver.switch_to.window()`

smithdc1:
I think there was a new marjor version of selenium recently. Maybe that introduced those changes.

It could be with splitting this PR into two changes. Likely the selenium ones will be easier to review and merge them the multiprocess ones.

jmacleod:
Ok, I will close this PR and split them.  Thank you!

PR #1840, opened 2022-04-16
feat(#490): Initial pyproject.toml creation

Atem18:


adamchainz:
We're not going to move the bulid system to flit. Setuptools works fine.

PR #1842, opened 2022-05-04
small fixes in tutorial part_4.rst

Khalmatov:
import By from selenium.webdriver.common.by and use custom method _switch_to_window instead of driver.switch_to.window


carltongibson (approved):
Great. Thanks.

PR #1854, opened 2022-06-03
Add subdomains of localhost to the default DEBUG mode allowed_hosts

ataylor32:
This will bring Channels in line with Django. See https://code.djangoproject.com/ticket/31010


carltongibson (commented):
Needs a test similar to https://github.com/django/django/pull/12132/files#diff-378adecd3fe6f72607f05dfc2a6fd2a259106e8beb502833230b61061a7412d4

Add to https://github.com/django/channels/blob/main/tests/security/test_websocket.py

PR #1862, opened 2022-06-27
tutorial/part_4: Fix old chrome_driver page to new

teacherSsamko:
- current link to chrome_driver will be deprecated soon [check out here](https://sites.google.com/a/chromium.org/chromedriver/downloads)


smithdc1 (approved):
Thanks -- this looks right. 

The deprecation notice on the old site is here https://sites.google.com/a/chromium.org/chromedriver/

PR #1870, opened 2022-07-10
Channels Projects now use Python 3.7+

smithdc1:
A small tweak to the readme. 

It seems to me, looking at the tox config that Channels now only supports Python 3.7+.


carltongibson (approved):
Great. Thanks.

PR #1905, opened 2022-08-19
Default code to 1005

caleb15:
See https://github.com/django/asgiref/issues/340. I'm assuming code is a `int`


carltongibson (commented):
Thanks @caleb15 — Any chance you could add a regression test for this?

PR #1910, opened 2022-08-25
Fix monospace formatting in Installation docs

johnthagen:
Prior to this fix:

Screen Shot 2022-08-25 at 7 57 04 AM


carltongibson (approved):
Oh, yes, good spot. Thanks.

PR #1918, opened 2022-09-16
Fix typo in the tutorial.

JulienPalard:
Found while testing channels 4 on my project.


carltongibson (approved):
Oh, yeah. We probably don't need that twice. Thanks. 😀

PR #1943, opened 2022-10-27
feat: fail on use ChannelsLiveServerTestCase when daphne is not installed

bertonha:
#1942 suggestion for fix the issue

### DIFF
--- a/channels/testing/live.py
+++ b/channels/testing/live.py
@@ -1,6 +1,10 @@
 from functools import partial
 
-from daphne.testing import DaphneProcess
+try:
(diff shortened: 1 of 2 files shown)


carltongibson (commented):
We should have a regression test here... 🤔

Perhaps an environment running the tests without Daphne installed (to make sure that works).

bertonha:
I think we have 2 options here:
- dont install daphne at all on the tests
- duplicate tests to run with daphne and without.

on the tests the class `ChannelsLiveServerTestCase` is not tested so far so daphne is not really needed at moment

Issue #458, opened 2017-01-04
Duplicate session key value error


denvaar:
Going through the getting started tutorial, I ran into the following error when creating a new socket and sending some text client-side. The app works as expected otherwise and this error doesn't appear to hinder the application. I am not sure why it is happening.

```
web_1    | [2017/01/04 04:41:29] WebSocket CONNECT /chat/ [172.17.0.1:55758]
db_1     | ERROR:  duplicate key value violates unique constraint "django_session_pkey"
db_1     | DETAIL:  Key (session_key)=(chna5e294d7179e6a194dfddf1e0e99a) already exists.
db_1     | STATEMENT:  INSERT INTO "django_session" ("session_key", "session_data", "expire_date") VALUES ('chna5e294d7179e6a194dfddf1e0e99a', 'MDExODI1NjViY2E5NzJmYjU0YmY1MjBkMTgwZjI4ZTA3ZDhiNTc3Mzp7fQ==', '2017-01-18T04:41:29.317408+00:00'::timestamptz)
```

Here is the Javascript that I ran:

```javascript
// Note that the path doesn't matter for routing; any WebSocket
// connection gets bumped over to WebSocket consumers
socket = new WebSocket("ws://" + window.location.host + "/chat/");
socket.onmessage = function(e) {
    alert(e.data);
}
socket.onopen = function() {
    socket.send("hello world");
}
// Call onopen directly if socket is already open
if (socket.readyState == WebSocket.OPEN) socket.onopen();
```

my consumers.py (the error appears to happen before any of this code is executed):
```python
from channels import Group
from channels.sessions import channel_session, enforce_ordering


@enforce_ordering(slight=True)
@channel_session
def ws_connect(message):
    room = message.content['path'].strip('/')
    message.channel_session['room'] = room
    Group('chat-{}'.format(room)).add(message.reply_channel)


@enforce_ordering(slight=True)
@channel_session
def ws_message(message):
    Group('chat-{}'.format(message.channel_session['room'])).send({
        'text': message['text'],
    })  


@enforce_ordering(slight=True)
@channel_session
def ws_disconnect(message):
    Group('chat-{}'.format(message.channel_session['room'])).discard(
        message.reply_channel)
```

Relevant info:

I'm running channels like so: `python server/manage.py runserver 0.0.0.0:8000`

- OSX 10.11.6
- Docker 1.12.3-beta29.3
- Chrome 55.0.2883.95
- Django 1.10.4
- asgi-redis 1.0.0
[...]

PR #1960, opened 2022-12-07
Upgrade to tox 4

adamchainz:
Copy of #458 for this repo.

tox 4 final is expected shortly: https://fosstodon.org/@gaborbernat/109428598380006519

tox-py no longer works with it, but that’s okay since the new `-f` factor option fills the need to run Python 3.X tests in the Python 3.X CI run. I documented this pattern in the package install instructions: https://github.com/adamchainz/tox-py/#installation . Plus, I’ve already changed all the projects I maintain to use it, for example: https://github.com/adamchainz/django-htmx/pull/296

Just like for channels - I've ported to more modern `setup.cfg` and `pyproject.toml` porject setup (which I initially thought was necessary for tox 4, but [it's not](https://github.com/django/daphne/pull/458#issuecomment-1340233154)). (aside: why no `long_description`? good to add one?).


carltongibson (approved):
Again, great. Please proceed when you're ready. 🎁 

> aside: why no long_description? good to add one?

Be my guest. 🦄

adamchainz:
Sorry rebase-and-merge got me and I’ve just added a bunch of tiny commits to `main` 😅

PR #1976, opened 2023-02-09
InMemoryChannelLayer improvements, test fixes

devkral:
Improvements to InMemoryChannelLayer
- honor expiry / don't block for eternity
- honor per channel capacities
- threadsafer cleaning design with atomic operations
- parallel sending

Bugfixes tests
- channel_layer is sometimes overwritten and the tests become bogus

### DIFF
--- a/channels/layers.py
+++ b/channels/layers.py
@@ -350,7 +350,11 @@ class InMemoryChannelLayer(BaseChannelLayer):
         if group in self.groups:
             for channel in self.groups[group].keys():
                 ops.append(asyncio.ensure_future(self.send(channel, message)))
-        await asyncio.wait(ops)
+        for send_result in asyncio.as_completed(ops):
+            try:
+                await send_result
+            except ChannelFull:
+                pass
 
 
 def get_channel_layer(alias=DEFAULT_CHANNEL_LAYER):
--- a/channels/layers.py
+++ b/channels/layers.py
@@ -50,7 +50,9 @@ class ChannelLayerManager:
         try:
             config = self.configs[name]["TEST_CONFIG"]
         except KeyError:
-            raise InvalidChannelLayerError("No TEST_CONFIG specified for %s" % name)
+            raise InvalidChannelLayerError(
+                "No TEST_CONFIG specified for %s" % name
+            )
         return self._make_backend(name, config)
 
     def _make_backend(self, name, config):
@@ -64,7 +66,9 @@ class ChannelLayerManager:
         try:
             backend_class = import_string(self.configs[name]["BACKEND"])
         except KeyError:
-            raise InvalidChannelLayerError("No BACKEND specified for %s" % name)
+            raise InvalidChannelLayerError(
+                "No BACKEND specified for %s" % name
+            )
         except ImportError:
             raise InvalidChannelLayerError(
                 "Cannot import BACKEND %r specified for %s"
@@ -168,10 +172,13 @@ class BaseChannelLayer:
     def valid_channel_names(self, names, receive=False):
         _non_empty_list = True if names else False
         _names_type = isinstance(names, list)
-        assert _non_empty_list and _names_type, "names must be a non-empty list"
+        assert (
+            _non_empty_list and _names_type
+        ), "names must be a non-empty list"
 
         assert all(
-            self.valid_channel_name(channel, receive=receive) for channel in names
+            self.valid_channel_name(channel, receive=receive)
+            for channel in names
         )
         return True
 
@@ -225,13 +232,14 @@ class InMemoryChannelLayer(BaseChannelLayer):
         # name in message
         assert "__asgi_channel__" not in message
 
-        queue = self.channels.setdefault(channel, asyncio.Queue())
-        # Are we full
-        if queue.qsize() >= self.capacity:
-            raise ChannelFull(channel)
-
+        queue = self.channels.setdefault(
+            channel, asyncio.Queue(maxsize=self.get_capacity(channel))
+        )
         # Add message
-        await queue.put((time.time() + self.expiry, deepcopy(message)))
+        try:
+            queue.put_nowait((time.time() + self.expiry, deepcopy(message)))
+        except asyncio.queues.QueueFull:
+            raise ChannelFull(channel)
 
     async def receive(self, channel):
         """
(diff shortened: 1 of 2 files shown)


devkral:
Note: I made an mistake with:
- honor expiry / don't block for eternity
This is the correct behavior for receive.
But maybe the receive method it should be put in a loop, in case the queue is expired
(see devkral/channels-multiprocessing for an example)

tests/test_inmemorychannel.py
```diff
@@ -62,7 +62,6 @@ async def test_multi_send_receive(channel_layer):
     """
     Tests overlapping sends and receives, and ordering.
     """
-    channel_layer = InMemoryChannelLayer()
```
bigfootjon, line 65: The recreation of `channel_layer` looks intentional to me, what do you mean when you say the tests become bogus?

devkral: nope, channel_layer is provided by a fixture. It makes absolutely no sense to create a channel_layer again.

(the fixture creates already fresh channel_layers)

channels/layers.py
```diff
@@ -279,19 +282,18 @@ def _clean_expired(self):
...
-        for group in self.groups:
-            for channel in list(self.groups.get(group, set())):
-                # If join time is older than group_expiry end the group membership
-                if (
-                    self.groups[group][channel]
-                    and int(self.groups[group][channel]) < timeout
-                ):
+        for channels in self.groups.values():
+
+            for name, timestamp in list(channels.items()):
```
bigfootjon, line 291: Why the extra `list()` call here?

channels/layers.py
```diff
@@ -329,22 +330,29 @@ async def group_discard(self, group, channel):
...
         assert self.valid_group_name(group), "Invalid group name"
         # Run clean
         self._clean_expired()
+
         # Send to each channel
-        for channel in self.groups.get(group, set()):
+        ops = []
+        if group in self.groups:
+            for channel in self.groups[group].keys():
+                ops.append(asyncio.ensure_future(self.send(channel, message)))
```
bigfootjon, line 352: Isn't it better to use `asyncio.create_task` instead of `asyncio.ensure_future`? Shouldn't we use the higher-level mechanism (as recommended in the docs)? Or is there a good reason to do it this way that I'm not seeing?

tests/test_inmemorychannel.py
```diff
@@ -62,7 +62,6 @@ async def test_multi_send_receive(channel_layer):
     """
     Tests overlapping sends and receives, and ordering.
     """
```
bigfootjon, line 64: Are you confident that the test suite covers all these changes? Are there any missing test-cases that we should add?

My particular areas of concern: the `del` -> `.pop` functionality, and `ChannelFull` handling, changes in queue capacity

channels/layers.py
```diff
@@ -225,13 +225,14 @@ async def send(self, channel, message):
...
         assert "__asgi_channel__" not in message
 
-        queue = self.channels.setdefault(channel, asyncio.Queue())
-        # Are we full
-        if queue.qsize() >= self.capacity:
-            raise ChannelFull(channel)
-
+        queue = self.channels.setdefault(
+            channel, asyncio.Queue(maxsize=self.get_capacity(channel))
+        )
```
bigfootjon, line 230: This changes the semantics around queue size right? What happens if the capacity is changed by the user while it is being used? I think that the new code would not honor that capacity change right?

channels/layers.py
```diff
@@ -242,14 +243,16 @@ async def receive(self, channel):
...
 
         # Do a plain direct receive
         try:
             _, message = await queue.get()
         finally:
             if queue.empty():
-                del self.channels[channel]
+                self.channels.pop(channel, None)
```
bigfootjon, line 255: Can you explain why this is an improvement? I'm not sure this is an improvement as it slightly hurts code readability IMO

(applies to all replacements of `del` in this PR)

bigfootjon (changes requested):
Thanks for the quick reply, sorry I'm just getting back to you.

This looks reasonable to me, I just have a few questions so I'm marking it as requested changes.

channels/layers.py
devkral: it is safer. If queue empty provides the wrong result, the code doesn't fail.

This can happen in async scenarios

channels/layers.py
devkral: because of channels.pop(name, None). Otherwise the dict size changes and python fails

channels/layers.py
devkral: good point. I am not sure about it. Is it really wished, that the capacity can change during runtime? Also it just affects not newly created channels.

tests/test_inmemorychannel.py
devkral: I agree, more tests would be better. I just invested no more time after getting no feedback long time

channels/layers.py
devkral: lowlevel is faster, but not sure about this.

bigfootjon: How much faster? Unless it's WAY faster I'm not sure it's worth going against the docs here.

tests/test_inmemorychannel.py
bigfootjon: Sorry about that, we're trying to do a better job of responding to issues/PRs. If you find time, please consider other test cases

channels/layers.py
bigfootjon: Can you provide a regression test to make sure this doesn't happen again in the future?

channels/layers.py
bigfootjon: Hm I suppose not. @carltongibson  do you have any thoughts?

channels/layers.py
devkral: it was a year ago, so most probably not worth the effort. Changed to create_task

tests/test_inmemorychannel.py
devkral: - I added a test for the race.
- changes in queue capacity are not happening (they are compiled statically in constructor) and if wanted we need to rewrite with a lock.
- ChannelFull is tested already

channels/layers.py
devkral: done

channels/layers.py
bigfootjon: I've thought about it and I can't think of any realistic scenario where this might cause problems, so I think the change is acceptable and we can document the change in the release notes

bigfootjon (approved):

PR #1998, opened 2023-04-05
Use format string for `room_name` in tutorial

johnthagen:
Since channels supports Python 3.7+, can use format strings to simplify string construction within the tutorial (which is a Python 3.6+ feature).


carltongibson (approved):
Grrr. Meh. OK, I guess. Thanks.

PR #2001, opened 2023-04-06
Update tutorial to use period style event types

johnthagen:
Something that confused me when going through the tutorial the first time was how exactly `chat_message()` method was called. This PR attempts to make that more clear.

It also moves the `type` to use `.` notation, which is more in-align with both how the ASGI spec names types as well as the guidance later in the docs: 

- https://channels.readthedocs.io/en/stable/topics/channel_layers.html#what-to-send-over-the-channel-layer

> Any message sent to that channel name - or to a group the channel name was added to - will be received by the consumer much like an event from its connected client, and dispatched to a named method on the consumer. The name of the method will be the type of the event with periods replaced by underscores - so, for example, an event coming in over the channel layer with a type of chat.join will be handled by the method chat_join.

I think this change will help users more naturally flow from the tutorial into what is considered typical best practice for naming event types.

### DIFF
--- a/docs/tutorial/part_2.rst
+++ b/docs/tutorial/part_2.rst
@@ -455,7 +455,7 @@ following code in ``chat/consumers.py``, replacing the old code:
 
             # Send message to room group
             async_to_sync(self.channel_layer.group_send)(
-                self.room_group_name, {"type": "chat_message", "message": message}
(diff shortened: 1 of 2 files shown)


carltongibson (approved):
OK, yep, agreed. 👍

PR #2017, opened 2023-06-14
Document `channels-auth-token-middlewares` third party package

johnthagen:
[`channels-auth-token-middlewares`](https://github.com/YegorDB/django-channels-auth-token-middlewares) fills an important role for providing authentication support out of the box for DRF applications using JWT authentication.

This question has been asked several times on SO:

- https://stackoverflow.com/questions/65297148
- https://stackoverflow.com/questions/43392889

Hopefully by promoting it in the community page, more users can discover it.


johnthagen:
CC @YegorDB for awareness

carltongibson (approved):
Yep, OK. Thanks!

PR #2091, opened 2024-04-15
Drop long deleted cookie_date

bigfootjon:
This was originally added in: https://github.com/django/channels/pull/1237

This was removed in https://github.com/django/django/commit/958a7b4ca69434d0145fd569cf007e21841bb36c (part of the 3.0 release)

I think we can drop the fallback now 😄


carltongibson (approved):
Nice. Yes.

PR #2101, opened 2024-06-06
Don't actually close DB connections during tests

bigfootjon:
This is effectively a combination of a modernization of https://github.com/django/channels/commit/9ae27cb835763075a81ba3823d20bd68f3ace46b and @adamchainz's minimal repro: https://github.com/django/channels/issues/1091#issuecomment-1294820937

Refs: https://github.com/django/channels/pull/2090
Closes: https://github.com/django/channels/issues/1091
Closes: https://github.com/django/channels/pull/1290

I fought with django for some time to try and convince it to use my overridden DB connection *just* inside these unit tests but it wasn't biting. I opted instead to just modify channels' global django settings

### DIFF
--- a/channels/testing/__init__.py
+++ b/channels/testing/__init__.py
@@ -1,5 +1,4 @@
-from asgiref.testing import ApplicationCommunicator  # noqa
-
+from .application import ApplicationCommunicator  # noqa
 from .http import HttpCommunicator  # noqa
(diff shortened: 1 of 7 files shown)


carltongibson (approved):
I think we should advance this, thanks @bigfootjon. 

My only consideration is to make sure we call out the change in the release notes for this.

PR #2103, opened 2024-06-12
Remove no more used deprecation message

sevdog:
This deprecation message usage was removed in 354284550204510ad02e5677a87605ccc4f228b7, thus there is no point in keeping it in the codebase.


carltongibson (approved):
Yep. Great. Thanks.

PR #2114, opened 2024-09-03
Fix handshake response headers

MRongM:
Fix handshake response headers with daphne server.

### DIFF
--- a/channels/generic/websocket.py
+++ b/channels/generic/websocket.py
@@ -53,6 +53,9 @@ class WebsocketConsumer(SyncConsumer):
         if headers:
             message["headers"] = list(headers)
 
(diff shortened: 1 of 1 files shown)


bigfootjon:
Can you explain what problem this solves?

MRongM:
The websocket client can't receive the handshake custom response headers.
I found that just setting the message["headers"] field did not work. Because the successHandshake method of the autobahn package that daphne relies on will determine whether the sub-protocol is a tuple.Then set headers based on the second field of this tuple.
You can check the link below:
https://github.com/crossbario/autobahn-python/blob/7bc85b34e200640ab98a41cfddb38267f39bc92e/autobahn/websocket/protocol.py#L2954C17-L2954C20

PR #2115, opened 2024-09-03
Update the Tutorial 

Lukas2019:
I tried to follow the tutorial but it was not working so I did a bit of research and found a solution. I updated the tutorial so that everyone else can follow the tutorial and get a working chat.

### DIFF
--- a/docs/tutorial/part_1.rst
+++ b/docs/tutorial/part_1.rst
@@ -81,6 +81,23 @@ following contents:
             urls.py
             wsgi.py
 
+Updating your settings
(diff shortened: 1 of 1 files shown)


bigfootjon:
The instructions you're adding here are already covered by other parts of the tutorial right?

1. https://channels.readthedocs.io/en/latest/tutorial/part_1.html#integrate-the-channels-library
2. https://channels.readthedocs.io/en/latest/tutorial/part_2.html#enable-a-channel-layer

Is this change necessary?

PR #2146, opened 2025-03-31
Restructured exceptions so exceptions within the backend are not hidden

wolph:
The current release of channels hides all `KeyError`s that can occur in backends. We should only catch the specific `KeyError` from parsing the settings, not from the import as well.

This pull request fixes that :)

### DIFF
--- a/channels/layers.py
+++ b/channels/layers.py
@@ -63,14 +63,18 @@ class ChannelLayerManager:
             )
         # Load the backend class
         try:
-            backend_class = import_string(self.configs[name]["BACKEND"])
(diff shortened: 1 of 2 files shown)


carltongibson (commented):
Hi @wolph — Do you think you would be able to add regression tests here, in `test_layers.TestChannelLayerManager`?

wolph:
@carltongibson I let openhands add some tests, they look good to me but to verify, can you allow the CI to run?

bigfootjon (approved):

carltongibson (approved):
OK, thanks @wolph 👍

PR #2162, opened 2025-06-08
Add ValidURLRouter to handle unmatched routes (#2147)

sprezz-arthur:
Follow-up from https://github.com/django/channels/issues/2147.

My attempt to solve the status code 500 on unmatched routes.
Solving it for http requests was quite straight-forward, as I could return 404 directly, but for websocket requests I could only manage to get a 403 by prematurely closing the connection before the handshake.

To make it non-breaking, I opted for extending `URLRouter` into `ValidURLRouter`, as some of the tests in `tests/test_routing.py` rely on having Exceptions raised.


sprezz-arthur:
Could we give it another go? 😅

PR #2164, opened 2025-06-11
fix django 5.2 compatibility

devkral:
Changes:

- fix django 5.2 compatibility
- add tests for ChannelsLiveServerTestCase
- add test project and use its settings
- fix hangup when the server failed to start


devkral:
sry, yes. I was asked a couple of weeks ago but had no time. Didn't checked the state properly.

But I fix some other issues too. We may should use my fix for the ChannelsLiveTestCase and his tests+sample project

PR #2179, opened 2025-08-08
Use signals to handle DB-close actions

sevdog:
This is a proposal to solve #2118.

I have resurrected the signals which were used in 1.x (in the old worker component) and put them in the consumer and `database_sync_to_async` lifecycle.

I added a context-manager in consumer to handle this (yet I feel that it is not the best solution) to disconnect-connect them.

Since #2118 does not get triggered when sqlite is used as DB backend, it need a database which is accessed through a socket, I would like to discuss the best way to put a test against a "real" database in the CI.

### DIFF
--- a/tests/sample_project/config/settings.py
+++ b/tests/sample_project/config/settings.py
@@ -70,6 +70,8 @@ DATABASES = {
         "NAME": "channels",
         "USER": "channels",
         "PASSWORD": "channels",
+        "HOST": "localhost",
+        "PORT": 5432,
     }
 }
 


carltongibson:
Hey @sevdog 

> Since https://github.com/django/channels/issues/2118 does not get triggered when sqlite is used as DB backend, it need a database which is accessed through a socket, I would like to discuss the best way to put a test against a "real" database in the CI.

Postgres is pretty easy to set up in GHA: 

```
jobs:
  build:
    runs-on: ubuntu-latest
    services:
      postgres:
        env:
          POSTGRES_USER: postgres
          POSTGRES_PASSWORD: postgres
          POSTGRES_DB: noumenal
        image: postgres:16.4
        ports: ["5432:5432"]
        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
```

Just initial take, I quite sceptical that this is the way forwards. It seems like a sledgehammer to crack a nut. 

From your comments on the issue, I wondered if just patching at the start of the process, rather than trying to do it in such a narrow context as we are might be more robust. 

> It would be good to have some kind of semaphore which tells to the channels.db package something like "we are running tests, let the connection status to be handled by test runner".

Crudely, yes, let's try setting a flag and no-oping on that. Does that resolve the issue. (In which case we can think about being more sophisticated.) 

cc @bigfootjon

carltongibson:
@sevdog I note this would require us to drop Django 4.2. That might be OK but I'm wondering if there isn't a smaller workaround you could apply. 

Is there a reliable reproduce, in the test project from the issue maybe?

PR #2196, opened 2025-11-14
Adds router resolver 404 exception

vladimirnani:
Motivation:
- Hard breaking in channels on non existent routes (because of bots/scanners) makes a lot of noise in logs analytics.

Contents:
- Add Django Resolver404 to unmatched routes
- ValueErrors will remain for misconfigurations and route not found exceptions become explicit
- Adds `RouterResolver404` as an abstraction over Django's `Resolver404`
- Keeps old `ValueError` exceptions handlers working for backwards compatibility

https://github.com/django/channels/issues/2147

### DIFF
--- a/channels/exceptions.py
+++ b/channels/exceptions.py
@@ -1,3 +1,6 @@
+from django.urls import Resolver404
+
+
 class RequestAborted(Exception):
     """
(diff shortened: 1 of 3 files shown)


vladimirnani:
No worries take your time. Thanks

PR #2199, opened 2025-11-21
Update Daphne docs, removing references to WhiteNoise

brianhelba:
As of [WhiteNoise 4.0](https://whitenoise.readthedocs.io/en/latest/changelog.html#id27) (released 2018-08-10), WhiteNoise supports running with Daphne. See [the fix in WhiteNoise](https://github.com/evansd/whitenoise/commit/ce74438f4de7e35aa922bf76e2159768708c6667) and [the original issue](https://github.com/evansd/whitenoise/issues/77).

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -62,11 +62,9 @@ take control of the ``runserver`` command. See :doc:`introduction` for more.
 
     Please be wary of any other third-party apps that require an overloaded or
     replacement ``runserver`` command. Daphne provides a separate
-    ``runserver`` command and may conflict with it. An example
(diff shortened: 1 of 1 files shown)


carltongibson (changes requested):
I think just removing the white noise reference entirely is better than replicating the configuration details here.

brianhelba:
That's fine, but note that Channels will break WhiteNoise if Channels is placed first. To be fair, it's generally true for any other package that overrides part of `runserver`.

I'll just make a PR over at WhiteNoise, to mention the threat from Channels on their docs, since they take extra steps to try and be compatible with multiple `runserver` overrides.

Issue #2208, opened 2025-12-11
ChannelsLiveServerTestCase keeps old ContentType cache between tests


johanneswilm:
When running multiple ChannelsLiveServerTestCase tests, the id values of the contenttypes change between test runs, yet the cache is not emptied. So `ContentType.get_for_model(X)` will get the instance from a previous test.

The only work-around I found was to add the following within the django view itself:

```py
if 'test' in sys.argv:
            ContentType.objects.clear_cache()
```

It is not enough to add this in the `setUp` or `tearDown` methods of the test, as the test runs in a separate thread and therefore has a separate cache. If someone can tell me how to, I'd very much like to move this to to the test code.

Tested with Django 5.2.9 and Django Channels 4.3.2.

johanneswilm:
Yeah, I guess I'll have to provide that. My code is all open source and publicly available [1], but it's big and bulky.

Is there some kind of minimal example channels test project that can easily be forked to create minimal reproduction examples?


[1] https://github.com/fiduswriter/fiduswriter/blob/main/fiduswriter/testing/selenium_helper.py#L171-L175

carltongibson:
Not really. But there is a live server set of tests in the test suite that maybe you can leverage

johanneswilm:
The problem seems to be in Channels 4.3.0. Downgrade to Django 5.1.0 and Channels 4.2.2 and the tests don't fail. Then upgrade to Channels 4.3.0 and they fail again.

carltongibson:
@johanneswilm Good work. Thanks. Let me play with it.

johanneswilm:
@carltongibson The change seems to have been introduced here: https://github.com/django/channels/commit/cd447717a84a8800081c437fc3a1ee4b2f616ae6 . Instead of ione sevrer process per test, there is now one server process for the entire class. And once the server process is running, it won't listen to regular cleanup commands as test and daphne run in different threads.

PR #2209, opened 2025-12-17
Add server command queue to ChannelsLiveServerTestCase to allow clean…

johanneswilm:
This is a possible solution for #2208 in combination with https://github.com/johanneswilm/django-channels-contenttype-test/pull/1


carltongibson:
No me neither currently 😜 Let me think about it. Thanks for the efforts. Really helpful

PR #2211, opened 2025-12-22
Removed dependency of `setup.cfg` from  project

p-r-a-v-i-n:
I think now it is safe to remove 'setup.cfg' 

i have produced the build and wheel on main  and on current branch ,and  I don't see differences (potentials )


carltongibson (commented):
@p-r-a-v-i-n Happy to take this if you could rebase it after #2220 

Thanks for the effort! 🏅

pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
     "asgiref>=3.9.0,<4",
 ]
 
-[project.optional-dependencies]
+[dependency-groups]
```
p-r-a-v-i-n, line 46: Do we want this in seperate PR or we are ok with here?
https://peps.python.org/pep-0735/

tox.ini
```diff
@@ -21,6 +21,7 @@ skip_install=true
 deps =
     black
     flake8
+    flake8-pyproject
```
p-r-a-v-i-n, line 26: This is for loading  config for flak8 from pyproject.toml

pyproject.toml
carltongibson: It's fine here. 👍

carltongibson: **BUT** it's not clear we should move everything here to dependency groups. 

In particular, `daphne` and `types` are still best seen as `extras` no? 

**And** can you scan the repo for docs/etc usages and make sure they're updated? 

e.g. https://channels.readthedocs.io/en/latest/contributing.html#how-do-i-get-started-and-run-the-tests

At that point I think it looks good.

p-r-a-v-i-n: Thank you so much Carlton to specifically spotting this. and you are correct here to raise this doubt.
IMO currently we shouldn't move `daphne` and `types` to dependency groups atleast for now.

>  can you scan the repo for docs/etc usages and make sure they're updated?

Yes I will make sure update every relevant place. Thank you :)

CONTRIBUTING.rst
```diff
@@ -25,16 +25,10 @@ Make sure the tests pass:
...
+    python -m pip install -e ".[daphne]"
     pytest
 
-.. note::
-   If you're using ``zsh`` for your shell, the above command will fail with a
-   ``zsh: no matches found: .[tests]`` error.
-   To fix this use ``noglob``::
-
-       noglob python -m pip install -e .[tests]
-
```
p-r-a-v-i-n, line 37: I removed this note because the note was specific to` .[tests],` and we will not use that syntax for tests. 

For extras like `daphne`, we use now `'.[daphne]'` which also avoids the zsh issue

PR #2220, opened 2026-04-20
Drop support for Django 4.2 and python 3.9

p-r-a-v-i-n:
check [here](https://docs.djangoproject.com/en/6.0/faq/install/#what-python-version-can-i-use-with-django:~:text=For%20example%2C%20Python%203%2E9%20security%20support%20ends%20in%20October%202025%20and%20Django%204%2E2%20LTS%20security%20support%20ends%20in%20April%202026%2E%20Therefore%20Django%204%2E2%20is%20the%20last%20version%20to%20support%20Python%203%2E9)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-        - "3.9"
         - "3.10"
         - "3.11"
         - "3.12"
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,14 +16,12 @@ classifiers =
     Operating System :: OS Independent
     Programming Language :: Python
     Programming Language :: Python :: 3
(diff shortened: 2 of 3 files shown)


carltongibson:
Looks about right @p-r-a-v-i-n. Thanks. 

We probably need to remove Python 3.9 from the GHA workflows too.

carltongibson (approved):
Let's have it. Thanks @p-r-a-v-i-n 🎁

p-r-a-v-i-n:
Hey @carltongibson , For curiosity I wanna know why do we still support django 5.1 ,  and i have seen this not in also channel but in other django packages too.

https://www.djangoproject.com/download/

carltongibson:
@p-r-a-v-i-n — Yep, we can drop 5.1 too. Fancy doing that?

p-r-a-v-i-n:
>  Fancy doing that?

yes

